PR #18, opened 2024-10-15
Trust self-signed certificates when using the eHX FHIR proxy

DanPaseltiner:
# PULL REQUEST

## Summary

The eHealth Exchange FHIR proxy currently uses a self-signed TLS certificate. Since we are no longer trusting self-signed certs globally in the application to restore our connection with eHX we need to configure trust them when making requests to eHX 

## Related Issue

Fixes #2578

## Acceptance Criteria

## Additional Information

## Checklist


robertandremitchell (approved):
looks good. seemed a tad slow when I tested with Cucumber Hill but otherwise worked

DanPaseltiner:
> looks good. seemed a tad slow when I tested with Cucumber Hill but otherwise worked

Using eHX adds more hops to responses are slower and if you used the Cerner server that one tends to be slower still because there is an auth process that happens as well.

PR #25, opened 2024-10-16
align customize query flow to design

fzhao99:
# PULL REQUEST

Apologies for the large PR!

## Summary
Breaks the query flow into the intended four step process by 

- Moving the customize query into a subflow of select-query, updating the mode type / constants as needed.
- Lifting / pushing down query response and value states into the component steps where it's needed, hopefully to simplify which components know about which states
- Moving the async fetch calls that used to be scattered around the flow into one place within `SelectQuery` / their own hooks file.
- Update tests to reflect the new stepped flow / copy changes, including splitting the tests up into separate files to help them run a bit faster

## Related Issue
After all the demo v0 pages are implemented independently, ensure that the pages link to one another according to [the user flow ](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/VIPER-Viewer-(Name-TBD)?node-id=4-6068&t=723PFItWMBWFEwQe-1) that Michelle put together

Fixes [#2663](https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi/2663)

### For design 
@mikang 
- Step 1 Patient discovery
    -  Added "step 1" to the title
    -  Color / spacing on the search form
    -  Misaligned from design: the "Advanced" link isn't in Figma right now but we mentioned we needed to have that as a technical constraint. Treated it like another section of the form, but happy to change as needed.

- Step 2 Patient Selection
    -  Added "step 2" to the title
    -  Added redirect to query selection

- [Step 3 Query Selection](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=17-8531&m=dev)
    -   Implemented the backlink to return to "Select patient"
    -  Added a loading state since the result query flow takes quite a while. 
    -  Also ported over the toast message from previous implementation where we went back to Patient Discovery from customize query. Let me know if still want this.
      Screenshot 2024-10-15 at 3 55 17 PM
    -  Noticed also for some of the use cases, the "apply changes" button that actually saves your selection is below the fold. Flagging since some users might do their selections and then hit "return", which won't actually save things.
Screenshot 2024-10-15 at 3 55 45 PM
    -  Noting that the "customize query" selection state will only maintain itself within the select query page despite what we talked about in Slack. Getting it to persist longer will be kinda complicated, so I didn't want it to block this work. Can have a followup ticket if we want to add that.   

- Step 4 Patient Record shouldn't have changed
     - The table widths in this page might need to be tweaked since there's horizontal scrolling on some default viewports


query-connector/e2e/alternate_queries.spec.ts
```diff
@@ -0,0 +1,81 @@
+import { test, expect } from "@playwright/test";
```
fzhao99, line 1: this whole page is just a copy-paste from the existing file to help parallelize the playwright runners

query-connector/e2e/load.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { test, expect } from "@playwright/test";
```
fzhao99, line 1: this whole page is just a copy-paste from the existing file to help parallelize the playwright runners

query-connector/src/app/CustomQuery.ts
```diff
@@ -75,7 +75,7 @@ export class CustomQuery {
         : "";
     this.medicationRequestQuery =
       rxnormFilter !== ""
-        ? `/MedicationRequest?subject=${patientId}&code=${rxnormFilter}&_include=MedicationRequest:medication&_include=MedicationRequest:medication.administration`
+        ? `/MedicationRequest?subject=${patientId}&code=${rxnormFilter}&_include=MedicationRequest:medication&_include=MedicationRequest:intended-performer`
```
fzhao99, line 78: For some reason, Public HAPI didn't like `medication.administration` as an includes param, which was breaking the inclusion of the medication requests for the patient. I did a bit of sparse googling and _think_ this is the closest of the provided options below from the error message but if someone knows better lmk and I'm happy to change it.

```
Invalid _include parameter value: &quot;MedicationRequest:medication.administration&quot;. Unknown search parameter &quot;medication.administration&quot; for resource type &quot;MedicationRequest&quot;. Valid search parameters for this search are: [encounter, intended-dispenser, intended-performer, medication, patient, requester, subject]
```

query-connector/src/app/query/components/CustomizeQuery.tsx
robertandremitchell, line 1: more minor, but we probably want the return here to Return to query selection, rather than return to patient

query-connector/src/app/query/components/selectQuery/SelectSavedQuery.tsx
```diff
@@ -0,0 +1,141 @@
...
+          <h3 className="margin-bottom-3">Health Care Organization (HCO)</h3>
+          <Select
+            id="fhir_server"
+            name="fhir_server"
+            value={fhirServer}
+            onChange={(e) => setFhirServer(e.target.value as FHIR_SERVERS)}
+            required
+            className={`${styles.queryDropDown}`}
+          >
+            Select HCO
```
robertandremitchell, line 103: do we want to remove this section of code if HCO has to move to SearchForm?

robertandremitchell (commented):
few minor quibbles, but looks great

query-connector/src/app/query/components/CustomizeQuery.tsx
fzhao99: good catch! will change this

query-connector/src/app/query/components/selectQuery/SelectSavedQuery.tsx
fzhao99: I'll let Dan and Michelle way in on this but I _think_ the idea is that the HCO might be needed here in the event that we want to scope a query for the selected patient into a specific HCO. Feel free to let me know if I'm wrong though

robertandremitchell (approved):
looks good. one minor thing we might want to consider:
when clicking `Search for patient` with zero fill-ins, it takes you to a page with a bunch of dummy data patients. We might want to disable the search for patient button unless some combination of name/MRN/phone/DOB are provided.

query-connector/e2e/load.spec.ts
m-goggins: Nice! Did you notice any difference by splitting it out?

m-goggins (approved):
I haven't been as close to this work but it looks really nice 😎 and was working well when I ran the changes locally

query-connector/e2e/load.spec.ts
fzhao99: Actually looking at the logs again it looks like we're just using the one worker, so don't think this did anything 😅

But hopefully this will help us into the future / a general positive organizational change

m-goggins: haha well at least it will make it easier to go through tests!

query-connector/src/app/CustomQuery.ts
m-goggins: Ah I didn't see this before during my original review. I can take a look at it tomorrow morning and see what I come up with

bamader: From what I could gather, the `administration` property seems more like how the dosage is administered. See the table here https://build.fhir.org/medicationrequest.html#bnr . Some of the examples given are taking an oral tablet, receiving a shot, etc. On the other hand, `intended-performer` is the individual who actually applies the medication to the patient (evidenced by the `Reference` property pointing to the patient, a physician, etc.). Unfortunately none of these actually relate to how the medication is taken--dispenser refers to the filling pharmacy, performer to the one giving the drug, etc. `encounter` _might_ be relevant as another event-type resource, but I don't love the fit of any of them :/

bamader (approved):
Looks great to me and works well locally! The medication.administration property is really annoying but I'm pretty tempted to open a follow up bug fix for it and let this PR land since it does so much to correctly align the workflow with designs.

DanPaseltiner:
> Noting that the "customize query" selection state will only maintain itself within the select query page despite what we talked about in Slack. Getting it to persist longer will be kinda complicated, so I didn't want it to block this work. Can have a followup ticket if we want to add that.

@fzhao99 persisting beyond the scope of the query at hand is beyond what we are hoping to do at this point. Next sprint we will begin tackling query building which will give users the ability to create and edit their queries not just the 6 we currently offer out of the box. Long way of saying you are totally right, but this limitation is fine for now.

query-connector/src/app/CustomQuery.ts
fzhao99: Thanks for the clarification Brandon! Will note that in the current cut of main medication requests seem to be broken because of this issue, so even without an exact match this will be an improvement over what exists. Would appreciate follow-ups to make this work as intended though!

DanPaseltiner:
I think we need to change the copy here 

image

As users will have already provided demographic data by this point. We could address in a separate ticket though.

@fzhao99 @mikang

query-connector/src/app/CustomQuery.ts
m-goggins: I looked into this a little more and I think we can update the query to look for this instead: `&_revinclude=MedicationAdministration:request`, which should pull back the Medication Administrations associated with the request.

mikang:
> I think we need to change the copy here
> 
> image
> As users will have already provided demographic data by this point. We could address in a separate ticket though.
> 
> @fzhao99 @mikang

Good catch @DanPaseltiner 

I think we can remove the supporting text and have just the bold text that says "Query customization successful." Thoughts?

mikang:
Update for visibility — Bob and I paired live, so all good to go from the design side once those changes are made! Mainly spacing

PR #40, opened 2024-10-21
Updating eslintrc

robertandremitchell:
# PULL REQUEST

## Summary

Adds back commented out code and fixing any code that is still flagged as problematic.

This was a little trickier to clean up existing code to be conformant, the biggest one is probably type guards, which I will concede is probably not the best handled, but that seemed to be what was recommended to clean out the `any`s in that situation.

## Related Issue

Fixes #2783

## Acceptance Criteria

## Additional Information

## Checklist


query-connector/tests-examples/demo-todo-app.spec.ts
fzhao99, line 1: should we just delete this file entirely?

query-connector/src/app/query-service.ts
```diff
@@ -233,24 +238,54 @@ export async function parseFhirSearch(
...
   }
 
   // Add resources to queryResponse
   for (const resource of resourceArray) {
-    const resourceType = resource.resourceType as keyof QueryResponse;
-    if (!queryResponse[resourceType]) {
-      queryResponse[resourceType] = [resource];
-    } else {
-      queryResponse[resourceType]!.push(resource);
+    if (isFhirResource(resource)) {
```
fzhao99, line 252: question for my understanding: what's the reason we need to switch case this out compared to what we used to have?

query-connector/tests-examples/demo-todo-app.spec.ts
robertandremitchell: I think we could, but I was honestly not sure if it was an in-progress or something we still needed.

query-connector/src/app/query-service.ts
robertandremitchell: I don't think we necessarily need to; however, what I was running into was that the any here
``` let resourceArray: any[] = [];```
was getting flagged.
When you just create the the FhirResource, it introduces a new error because patient/observation/encounter can have different shapes.

The switch/case is definitely very very clunky, but I could not figure out another way to get rid of that error that also moved away from `any`.

query-connector/tests-examples/demo-todo-app.spec.ts
fzhao99: I think this comes as[ an example out of the box from Playwright](https://github.com/microsoft/playwright/blob/main/examples/todomvc/tests/integration.spec.ts), so I think we can remove it

fzhao99:
lgtm! will give an approval once Brandon's PR goes in since we're locking main

fzhao99 (approved):

PR #43, opened 2024-10-22
Refactor new data model

bamader:
# PULL REQUEST

## Summary

This PR refactors our code base to leverage the new value set and concept data models that Marcelle developed as part of her spike work. Refactoring is done throughout the front and back end, fixing how we execute queries as well as display results.

## Related Issue

Fixes https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi/2789

### DIFF
--- a/query-connector/src/app/database-service.ts
+++ b/query-connector/src/app/database-service.ts
@@ -1,11 +1,11 @@
 "use server";
 import { Pool, PoolConfig, QueryResultRow } from "pg";
 import { Bundle, OperationOutcome } from "fhir/r4";
-import { ValueSetItem, valueSetTypeToClincalServiceTypeMap } from "./constants";
 import { encode } from "base-64";
+import { ValueSet } from "./constants";
 
 const getQuerybyNameSQL = `
-select q.query_name, q.id, qtv.valueset_id, vs.name as valueset_name, vs.author as author, vs.type, qic.concept_id, qic.include, c.code, c.code_system, c.display 
+select q.query_name, q.id, qtv.valueset_id, vs.name as valueset_name, vs.version, vs.author as author, vs.type, vs.dibbs_concept_type as dibbs_concept_type, qic.concept_id, qic.include, c.code, c.code_system, c.display 
   from query q 
   left join query_to_valueset qtv on q.id = qtv.query_id 
   left join valuesets vs on qtv.valueset_id = vs.id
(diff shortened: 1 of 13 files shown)


query-connector/src/app/database-service.ts
```diff
@@ -48,45 +48,48 @@ export const getSavedQueryByName = async (name: string) => {
...
-  // Mapping is established in TCR, so follow that convention
-  let valuesetFilters = valueSetTypeToClincalServiceTypeMap[type];
-  const results = vsItems.filter((vs) =>
-    valuesetFilters.includes(vs.clinicalServiceType),
-  );
-  return results;
-};
+  // Create groupings of rows (each of which is a single Concept) by their ValueSet ID
+  const vsIdGroupedRows = rows.reduce((conceptsByVSId, r) => {
+    if (!conceptsByVSId[r["valueset_id"]]) {
```
fzhao99, line 62: nit: could we do `if(!(r["valueset_id"] in conceptsByVSId))` here? Don't think this is very likely, but in the off chance that somehow the dictionary values get populated with something falsy, this might not behave like expected. The `in` keyword is made for cases like this specifically

query-connector/src/app/query/components/CustomizeQuery.tsx
```diff
@@ -99,6 +112,9 @@ const CustomizeQuery: React.FC<CustomizeQueryProps> = ({
       (item) => ({
         ...item,
         include: checked, // Set all items in this group to checked or unchecked
```
fzhao99, line 114: I think as it currently is this needs to be `includeValueSet` in order for this to work properly, since we have a different include type bool at the valueset vs concept level. In line 88 of `queryHooks`, we're checking `includeValueSet`, not include, which I think is causing the filtering to break.

In my mind though there's no reason we need different properties at the two levels: we should have them be the same to reduce the cognitive complexity of what consitutes "include". Kknowing it's quite a bit of work (😬), what do we think of changing this at the object level rather than having to remember the difference between the two?

query-connector/src/app/query/components/selectQuery/queryHooks.ts
```diff
@@ -85,7 +85,7 @@ export async function fetchQueryResponse(p: {
     p.setIsLoading(true);
     const queryResponse = await UseCaseQuery(
       newRequest,
-      p.queryValueSets.filter((item) => item.include),
+      p.queryValueSets.filter((item) => item.includeValueSet),
```
fzhao99, line 88: Do we need to do any filtering here at the concept level? Right now, it seems we'd just filter by the upper level valueset, which unless I'm missing something doesn't seem to be granular enough for individual concept-level filtering

fzhao99 (commented):
Great work on tackling such a thorny slice of the codebase! Left a couple of comments

(and hooray for no merge conflicts!)

query-connector/src/app/query/components/CustomizeQuery.tsx
fzhao99: Relatedly, I think we need to add an e2e to check the whole customize query flow since that's like, a core part of the current app. Made a ticket for that [here](https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/dibbs-query-connector/44)

query-connector/src/app/database-service.ts
bamader: sure, seems fine to change

query-connector/src/app/query/components/CustomizeQuery.tsx
```diff
@@ -66,22 +66,35 @@ const CustomizeQuery: React.FC<CustomizeQueryProps> = ({
   ).length;
```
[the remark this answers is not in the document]
bamader: Good catch, I updated the counts in the individual accordion headers but missed the count at the top of the page overall.

query-connector/src/app/query/components/CustomizeQuery.tsx
bamader: Could you elaborate on what you mean by the filtering is breaking? Visually, everything looks like it's working as expected so not really sure what you mean.

Re: two levels of include: that's something Marcelle put in her spike work, that both `Concept` and `ValueSet` should have include properties. I can't seem to find the reasoning for it, but given the depth she went on her research I'd probably prefer to leave this as-is, at least until she's back. Given that we're also going to have a discussion Thursday on DB schema, I would likely opt to not change the include structure in this PR, especially since it was ~3 days of work and I'd like to get it merged soon so we can capture the changes and keep everything up to date. Conceptually, I think the idea is that we include a value set if at least one concept in that value set is included in a query, so if that is the rationale then `includeValueSet` becomes a convenience property rather than something strictly "necessary."

bamader: Given my theoretical explanation of the rationale for two includes, I think the right approach would actually be to set both properties to checked.

query-connector/src/app/query/components/selectQuery/queryHooks.ts
bamader: I _think_ this should be okay as is? Conceptually, if my reasoning for the relationship between `include` and `includeValueSet` is correct, then this should still capture everything that a user has specified at least one concept for. I think this could equivalently be an inclusion of any value set for which at least one member of their `concept` list is set to include.

bamader:
(There actually were merge conflicts unfortunately, the refactor of the `SelectQuery` meant that I had to refactor the new changes this morning. Got it done rather quickly at least.)

query-connector/src/app/query/components/CustomizeQuery.tsx
bamader: LOL ignore everything I just wrote then

bamader: Ah excellent, that helps a lot. I'll dig into this and see if changing the includes helps. Thanks!

bamader:
All comments addressed and filtering should now be working as intended

bamader:
Concept filtering now appropriately in place

fzhao99 (approved):
thanks for going a couple rounds on this! lgtm!

PR #46, opened 2024-10-22
do some ts magic to get around the switch setup

fzhao99:
Adds a few utility classes to get around Typescript union type wonkiness

### DIFF
--- a/query-connector/src/app/query-service.ts
+++ b/query-connector/src/app/query-service.ts
@@ -14,22 +14,20 @@ import { CustomQuery } from "./CustomQuery";
 import { GetPhoneQueryFormats } from "./format-service";
 import { formatValueSetItemsAsQuerySpec } from "./format-service";
 
-type SuperSetQueryResponse = {
-  [R in FhirResource as R["resourceType"]]?: R[];
-};
-
 /**
  * The query response when the request source is from the Viewer UI.
  */
-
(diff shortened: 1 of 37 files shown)


query-connector/src/app/query-service.ts
```diff
@@ -297,14 +257,19 @@ export async function parseFhirSearch(
  * @param response - The response from the FHIR server.
  * @returns - The array of resources from the response.
  */
-export async function processResponse(
+export async function processFhirResponse(
   response: fetch.Response,
-): Promise<unknown[]> {
-  let resourceArray: unknown[] = [];
+): Promise<FhirResource[]> {
+  let resourceArray: FhirResource[] = [];
```
fzhao99, line 263: @robertandremitchell pushed the type checking from where you had it down a level to where we're making contact against the relevant FHIR response, which feels like the part of the stack where we'd check for it. RN just logging out a mismatch in the event that we have an issue rather than "properly" handling the issue based on the scope of the ticket, but lmk if you have any issues here

robertandremitchell: I think that makes sense. I had tinkered with moving it yesterday but it felt like it was getting more complex. this definitely simplifies it a lot, which is great!

robertandremitchell (approved):
lgtm!

Issue #44, opened 2024-10-22
Add an e2e to cover the customize query flow


fzhao99:
Right now we don't have one and we should since it's a core part of the app. 

Should cover both valueset level and concept level filtering

PR #53, opened 2024-10-23
Add e2e for customize query

bamader:
# PULL REQUEST

## Summary
This PR adds end to end testing for the customize query workflow. Various facets are tested, including selective valueset filtering, group checking/unchecking, and selecting/deselecting all.

## Related Issue
Fixes #https://github.com/CDCgov/dibbs-query-connector/issues/44

### DIFF
--- a/query-connector/e2e/customize_query.spec.ts
+++ b/query-connector/e2e/customize_query.spec.ts
@@ -104,9 +104,6 @@ test.describe("querying with the Query Connector", () => {
   test("customize query select / deselect all filters whole DibbsConceptType, across tabs", async ({
     page,
   }) => {
-    // await expect(
-    //   page.getByText("249 labs found, 4 medications found, 104 conditions found."),
-    // ).toBeVisible();
     await page.getByRole("link", { name: "Labs" }).click();
     await page.getByRole("button", { name: "Deselect all labs" }).click();
 
@@ -177,7 +174,6 @@ test.describe("querying with the Query Connector", () => {
     for (let i = 1; i < 6; i++) {
       const row = obsRows.nth(i);
       const typeText = await row.locator("td").nth(1).textContent();
-      console.log(typeText);
       const presentKey = acceptableSdohKeywords.find((key) =>
         typeText?.toLowerCase().includes(key),
       );
--- a/query-connector/e2e/customize_query.spec.ts
+++ b/query-connector/e2e/customize_query.spec.ts
@@ -1,0 +1,193 @@
+// @ts-check
+
+import { test, expect } from "@playwright/test";
+import { TEST_URL } from "../playwright-setup";
+import { PAGE_TITLES } from "@/app/query/stepIndicator/StepIndicator";
+
+import { TEST_PATIENT, TEST_PATIENT_NAME } from "./constants";
+
+test.describe("querying with the Query Connector", () => {
+  // Start every test by navigating to the customize query workflow
+  test.beforeEach(async ({ page }) => {
+    await page.goto(TEST_URL);
+    await page.getByRole("button", { name: "Go to the demo" }).click();
+
+    // Check that the info alert is visible and contains the correct text
+    const alert = page.locator(".custom-alert");
+    await expect(alert).toBeVisible();
+    await expect(alert).toHaveText(
+      "This site is for demo purposes only. Please do not enter PII on this website.",
+    );
+    await expect(
+      page.getByRole("heading", { name: PAGE_TITLES["search"], exact: true }),
+    ).toBeVisible();
+
+    await page.getByRole("button", { name: "Fill fields" }).click();
+    await page.getByLabel("First Name").fill(TEST_PATIENT.FirstName);
+    await page.getByLabel("Last Name").fill(TEST_PATIENT.LastName);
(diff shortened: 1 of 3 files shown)


query-connector/e2e/customize_query.spec.ts
```diff
@@ -0,0 +1,193 @@
+// @ts-check
+
+import { test, expect } from "@playwright/test";
+import { TEST_URL } from "../playwright-setup";
+import { PAGE_TITLES } from "@/app/query/stepIndicator/StepIndicator";
+
+import { TEST_PATIENT, TEST_PATIENT_NAME } from "./constants";
```
m-goggins, line 7: I'm not sure when this got added but definitely makes it easier to write tests!

query-connector/e2e/customize_query.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+    await expect(
+      page.getByRole("heading", { name: PAGE_TITLES["search"], exact: true }),
+    ).toBeVisible();
+
+    await page.getByRole("button", { name: "Fill fields" }).click();
+    await page.getByLabel("First Name").fill(TEST_PATIENT.FirstName);
+    await page.getByLabel("Last Name").fill(TEST_PATIENT.LastName);
+    await page.getByLabel("Date of Birth").fill(TEST_PATIENT.DOB);
+    await page.getByLabel("Medical Record Number").fill(TEST_PATIENT.MRN);
+    await page.getByLabel("Phone Number").fill(TEST_PATIENT.Phone);
```
m-goggins, line 30: Do we have to fill these fields if you already clicked "Fill Fields" on line 25?

query-connector/e2e/customize_query.spec.ts
```diff
@@ -0,0 +1,193 @@
...
+        .getByRole("table")
+        .filter({ hasText: "I do not have housing" })
+        .getByRole("row"),
+    ).toHaveCount(6);
+    const acceptableSdohKeywords = [
+      "history",
+      "narrative",
+      "housing",
+      "pregnancy",
+      "with anonymous partner",
```
m-goggins, line 176: This might be out of scope for the ticket (or honestly over optimization at this stage), but what do you think about creating a customized query constant like we do for `TEST_PATIENT`? We could add key words like this as well as the expected medications, etc. I'm not sure if it's easier to update a constant or just update the tests at this point but thought I'd throw it out there.

m-goggins (approved):
Left a few non-blocking comments, but overall looks good!

query-connector/e2e/customize_query.spec.ts
bamader: I thought the same thing! Was so glad to see it just appear haha. Might be harder to use once we support multiple default or test patients like we used to, but for now it's absolutely great.

query-connector/e2e/customize_query.spec.ts
bamader: Good question 🤔 . I just copied this setup stuff from the initial part of the regular query_spec tests, but I'd imagine not actually. As long as that test patient remains the same we shouldn't need them.

robertandremitchell (approved):
looks good!

i imagine some of the stuff at the beginning, like navigating to the searchform page, is carry-over from prior tests. I wonder if we need it here, or if we could just directly to the /query endpoint. If/when we do start to change the copy of the home page, it'll just mean having to update this, when in theory it has nothing to really do with our customize query code.

I also ran into some issues running the tests locally, but this seems to be a known confusing issue, so 🤷🏽

query-connector/e2e/customize_query.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+        .filter({ hasText: "doxycycline hyclate 100 MG" })
+        .getByRole("row"),
+    ).toHaveCount(2);
+  });
+
+  test("customize query select / deselect all filters whole DibbsConceptType, across tabs", async ({
+    page,
+  }) => {
+    // await expect(
+    //   page.getByText("249 labs found, 4 medications found, 104 conditions found."),
```
fzhao99, line 108:
```suggestion
```
did we mean to clean this comment up?

query-connector/e2e/query_workflow.spec.ts
```diff
@@ -62,23 +62,27 @@ test.describe("querying with the Query Connector", () => {
...
-    ).not.toBeVisible();
-
-    await page.getByText("Return to Select query").click();
+    // Update as the unlucky dev: Chromium now has issues with this check present
+    // for chlamydia, as well as issues in another test checking for non-zero values.
+    // Other browsers have no problem navigating both checks as they stand.
+    // Update to the update: the connection issue has broadened to include webkit
+    // and seems to apply with no rhyme or reason to use case. Commenting this out
+    // because it no longer acts as a sanity check but leaving it here as a warning
+    // to others.
```
fzhao99, line 77: 😭

fzhao99 (approved):
lgtm! sucks that you ran into whatever playwright edge case that bit me a few weeks ago, but glad things are working in CI

query-connector/e2e/customize_query.spec.ts
bamader: Yep, cleaned

PR #63, opened 2024-10-25
Rework name regex on patient search page

### LINKED ISSUE #56: Rework name regex on patient search page
### Why are we doing this?

The regex on the [first](https://github.com/CDCgov/dibbs-query-connector/blob/732bc72efc9552d5b0395c211e171d70863824ff/query-connector/src/app/query/components/SearchForm.tsx#L196) / [last](https://github.com/CDCgov/dibbs-query-connector/blob/732bc72efc9552d5b0395c211e171d70863824ff/query-connector/src/app/query/components/SearchForm.tsx#L215) name form fields of the patient search form are of type /g^[A-Za-z ]+$/, which invalidates things like apostrophes and em dashes that might be valid in patient names. We'll need to discuss what a better set of regex would be (or if we should be sanitizing the values outside of this frontend form field in the first place) and implement accordingly

image

### Action Requested

Do a bit of research for how we should be sanitizing name input for frontend form fields
Change / remove the regex accordingly

### Acceptance Criteria

Patient form field accepts a reasonable (up to the dev to determine what reasonable means!) set of patient names, which would include dashes / apostrophes etc. that might show up in a name

### Additional Context

_No response_

katyasoup:
# PULL REQUEST

## Summary
- Update regex for first name/last name to allow letters, hyphens, apostrophes, spaces, periods, and diacritics.
- Add `title` to `TextInput`s to display validation rule on tooltip message

## Related Issue

Fixes #56 

## Acceptance Criteria
_Patient form field accepts a reasonable (up to the dev to determine what reasonable means!) set of patient names, which would include dashes / apostrophes etc. that might show up in a name_

Allowed characters were chosen based on [USWDS Guidance](https://designsystem.digital.gov/patterns/create-a-user-profile/name/#guidance-2) and [HealthCare.gov](https://www.healthcare.gov/create-account) validation rules/messaging

Some names that should pass validation:
- A'ja Wilson
- Björk Guðmundsdóttir
- Julia Louis-Dreyfus
- Magdalena Carmen Frida Kahlo y Calderón
- St. Joan of Arc

Some names that should fail validation:
- Queen Elizabeth the 2nd
- Dwayne "The Rock" Johnson
- Panic! At the Disco

When an input fails validation, the tooltip should now display details on acceptable input characters like so:

Screenshot 2024-10-25 at 15 14 29

## Additional Information

There's an argument to be made (see [USWDS Usability Guidance](https://designsystem.digital.gov/templates/form-templates/name-form/#usability-guidance)) that we shouldn't restrict characters at all; however, this PR's changes align with current usage on various other HHS sites.

The browser tooltip for error handling isn't great for a11y either, so I'd recommend some future work on inline error styling a là the [CMS Design System](https://design.cms.gov/patterns/Forms/error-validation/?theme=core)

## Checklist

### DIFF
--- a/query-connector/src/app/query/components/SearchForm.tsx
+++ b/query-connector/src/app/query/components/SearchForm.tsx
@@ -79,7 +79,7 @@ const SearchForm: React.FC<SearchFormProps> = ({
     [setUseCase],
   );
 
-  const nameRegex = "^[A-Za-z\u00C0-\u024F\u1E00-\u1EFF\ \-\'\.]+$"
+  const nameRegex = "^[A-Za-z\u00C0-\u024F\u1E00-\u1EFF\-\'\.\ ]+$"
   const nameRuleHint = "Enter a name using only letters, hyphens, apostrophes, spaces, or periods."
 
   async function HandleSubmit(event: React.FormEvent<HTMLFormElement>) {
--- a/query-connector/src/app/query/components/SearchForm.tsx
+++ b/query-connector/src/app/query/components/SearchForm.tsx
@@ -79,6 +79,9 @@ const SearchForm: React.FC<SearchFormProps> = ({
     [setUseCase],
   );
 
+  const nameRegex = "^[A-Za-z\ \-\'\.]+$"
+  const nameRuleHint = "Enter a name using only letters, hyphens, apostrophes, spaces, or periods."
+
   async function HandleSubmit(event: React.FormEvent<HTMLFormElement>) {
     event.preventDefault();
     if (!fhirServer) {
@@ -191,9 +194,10 @@ const SearchForm: React.FC<SearchFormProps> = ({
               </Label>
               <TextInput
                 id="firstName"
+                title={nameRuleHint}
                 name="first_name"
                 type="text"
-                pattern="^[A-Za-z ]+$"
+                pattern={nameRegex}
                 value={firstName}
                 onChange={(event) => {
                   setFirstName(event.target.value);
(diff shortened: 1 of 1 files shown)


fzhao99 (approved):
appreciate the thoughtfulness / thoroughness of this implementation Katie! Lgtm. 

Hopefully we can [figure out what's wrong with our e2e's](https://skylight-hq.slack.com/archives/C05BEG226RZ/p1729882981695179) so we can get this merged.  Happy to help troubleshoot as needed!

query-connector/src/app/query/components/SearchForm.tsx
```diff
@@ -79,6 +79,10 @@ const SearchForm: React.FC<SearchFormProps> = ({
     [setUseCase],
   );
 
+  const nameRegex = "^[A-Za-z\u00C0-\u024F\u1E00-\u1EFF\\-'. ]+$";
```
m-goggins, line 82: Is this regex based on the USWDS standard? I'm wondering what your thoughts are on aligning more closely with the [FHIR r4 standard for strings](https://www.hl7.org/fhir/r4/datatypes.html), which has a more permissive regex, `[ \r\n\t\S]+`. 

Of course, the fhir regex is  "informative, not normative" so it seems we have some leeway with how we implement.

m-goggins (approved):
Looks good to me! I like the addition of the constants to make it clearer what regex is being applied to the field. 

I left a comment for your consideration about aligning with the FHIR r4 standard in addition to thinking about the USWDS/Healthcare.gov standards. It looks like the FHIR one is mostly a suggestion but wanted to bring it to your attention just in case.

query-connector/src/app/query/components/SearchForm.tsx
katyasoup: USWDS specifically calls out `hyphens, apostrophes, and blank spaces` as well as `"a rich array of diacritics, accents, and alternative characters"`, but then elsewhere advises `"Don’t restrict characters."` 😅 I think I'll leave it as-is for now but definitely happy to revisit if we decide we want to align more closely with r4!

PR #88, opened 2024-10-29
Adding Category data to Conditions table

### LINKED ISSUE #68: Determine how to store category information for conditions
Dan mentioned we got this information back from APHL. We need to determine how we store this information / how we expose this from the backend for rendering

robertandremitchell:
# PULL REQUEST

## Summary

This adds a new column to the `conditions` table, `category`. This is an organizational structure that we need to organize our conditions to display like so:  https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=475-18708&node-type=frame&t=Lcg6TdahnUvYPBsR-0

These are the categories we would end up with:
Screenshot 2024-10-29 at 1 47 36 PM

the three missing are ones we added. Here are my thoughts on how to handle:
* Social Determinants of Health, which I believe we can remove from the DB given convo today
* `Cancer (Leukemia)`, which, probably fairly reasonably, could be coded to coded to be included in the `Cancer` category.
* Newborn Screening, which could be `Birth Defects and Infant Disorders`? I don't see an immediately obvious one we would nest Newborn Screening under, otherwise. I can see an argument for it being standalone.

a few backend considerations: right now, we are just using INSERT statements to create the temporary `category_data` table that needs to be merged with `conditions`.

Other ideas:
* this data in theory is only necessary for the display of data on the custom query page. We could just have a small script that transforms this into a JSON we reference to functionally display this data independent of the db.
* we could read the data from an S3 bucket. This is probably a lot more work, but if we were uncomfortable committing this file as a standalone file to the repo, or if it introduces Terraform complications, that might provide some balance.
## Related Issue

Fixes #68 

## Acceptance Criteria

## Additional Information

## Checklist


query-connector/public/Conditions_and_Categories_20241024.csv
DanPaseltiner, line 1: I think the should probably put this file some where else as all of the other artifact in this directory are served publicly to end users in the UI. We probably aren't looking to provide direct access to this file

query-connector/flyway/sql/V01_07__add_conditions_category.sql
```diff
@@ -0,0 +1,34 @@
+-- 1. Add the new 'category' column to the condition table
+ALTER TABLE conditions
+ADD COLUMN category VARCHAR(255);
```
DanPaseltiner, line 3: We should probably use `TEXT` instead of `VARCHAR(255)`

[https://www.sprinkledata.com/blogs/postgresql-text-vs-varchar-choosing-the-right-data-type-for-your-database](url)

query-connector/public/Conditions_and_Categories_20241024.csv
robertandremitchell: makes sense. I think we'd need it in `public` (or at least what I read it would be easier in `public`) if any other parts of the code needed to interact with the CSV. but I don't think we should need to.

query-connector/flyway/sql/V01_07__add_conditions_category.sql
robertandremitchell: updated!

bamader (commented):
Looks fine to me as is, though I would probably be in the camp of just using INSERT statements to make that temporary table. I have no problem exposing this file publicly, but we probably aren't looking to do so. Mounting and S3 both seem like overkill to me, but that's more a stylistic and preference thing, I think--I would rather our DB product not establish the pattern of bringing in externally mounted volumes. I think the seeding/creation should be as self-contained as possible. I don't feel strongly either way, and I don't think anything here is blocking though!

robertandremitchell:
> Looks fine to me as is, though I would probably be in the camp of just using INSERT statements to make that temporary table. I have no problem exposing this file publicly, but we probably aren't looking to do so. Mounting and S3 both seem like overkill to me, but that's more a stylistic and preference thing, I think--I would rather our DB product not establish the pattern of bringing in externally mounted volumes. I think the seeding/creation should be as self-contained as possible. I don't feel strongly either way, and I don't think anything here is blocking though!

I'm torn, I think I lean toward a bunch of `INSERT` statements also being fine. I think I'd probably want to add a script that can parse this CSV to write those `INSERT`s each time a new eRSD drops for some future-proofing (or, preferably, convince APHL to surface this data in the eRSD...).

happy to look into that option if we think it's cleaner.

DanPaseltiner:
@bamader definitely agree that ultimately all database creation should be handled in a single place (the work you and @m-goggins are doing). I totally forgot about this when reviewing. I'd propose that we move this along so that @robertandremitchell and @fzhao99 aren't blocked on their query building work, but eventually have this along with all other seeding handled at the app layer.

bamader:
@DanPaseltiner That sounds good to me--approving with the idea that we revisit the conversation during broader DB creation!

bamader (approved):

robertandremitchell:
> @bamader definitely agree that ultimately all database creation should be handled in a single place (the work you and @m-goggins are doing). I totally forgot about this when reviewing. I'd propose that we move this along so that @robertandremitchell and @fzhao99 aren't blocked on their query building work, but eventually have this along with all other seeding handled at the app layer.

Makes sense to me. I've created a new issue in the backlog so that we don't lose track of that work (https://linear.app/skylight-cdc/issue/QUE-43/refactor-conditionscategory-migration), but I will work on getting this merged shortly (darn CI tests). 

Thank you both for your review!

fzhao99 (approved):
👍 

Agreed that we should probably avoid the mounting / external host route given the (relatively) static nature of this data. Also lean towards just doing it the brute force `INSERT` way, but obviously if we could get it through an API call that'd be best

robertandremitchell:
requesting a re-review since switching to `INSERT` over CSV. everything else should be the same.

fzhao99 (approved):

Issue #72, opened 2024-10-28
Add insertion for `query_to_valueset`


fzhao99:
Any concepts associated with the query is coming from a valueset in this pass. In other words, at this stage, we don't have to support “orphan” concepts in queries that don't have parent valuesets.

PR #89, opened 2024-10-29
define backend query building insert function

### LINKED ISSUE #70: Define backend query creation spec and functions
What's the contract between frontend user input and what the backend needs for query building? 

This is “shape of the data that comes from the frontend once we are done editing / adding to the new query and hit Back to my queries”: from [this page](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=475-32251&t=nCvdKQwvRKVyZnv4-4) to [this page](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=475-20613&t=nCvdKQwvRKVyZnv4-4)

fzhao99:
# PULL REQUEST
Backend insertion function needed for query building.

## Summary
You can visit and see the below changes at `{basePath}/queryBuilding` on [this test branch](https://github.com/CDCgov/dibbs-query-connector/tree/bob/query-insertion-test)

https://github.com/user-attachments/assets/6f9890de-8db1-41df-83ce-11da857a1634

- Sketches out the backend query building insertion function, including cascading join insertions for the `query_to_valuesets` 
- Defines input param type to for the frontend to send to the backend
     - Will sync this into the work being done for database creation in the event that the two types are similar enough to be merged   
- Slightly refactor the `ValueSet` type to include external/OID for future use - discussion of this addition / rework [here](https://skylight-hq.slack.com/archives/C05BEG226RZ/p1730297408883749)
- Refactor some frontend fetching logic to be a bit more generic from another component for future reuse

## Related Issue
Fixes #70 and begins implementation of #72  

## Acceptance Criteria

Builds out a test page at `/queryBuilding` that just accepts the query name and then inserts a custom query with a hardcoded group of valuesets to make sure the insertion flow works. Will build off this UI as we iteratively build it out to match the designs.

Will hide this from the UI navigation until it's ready

### DIFF
--- a/query-connector/src/app/constants.ts
+++ b/query-connector/src/app/constants.ts
@@ -319,6 +319,7 @@ export interface ValueSet {
   valueSetId: string;
   valueSetVersion: string;
   valueSetName: string;
+  valueSetExternalId?: string;
   author: string;
   system: string;
   ersdConceptType?: string;
--- a/query-connector/src/app/constants.ts
+++ b/query-connector/src/app/constants.ts
@@ -81,6 +81,7 @@ export const FhirServers = [
   "JMC Meld: Direct",
   "JMC Meld: eHealthExchange",
   "Public HAPI: Direct",
+  "Local e2e HAPI Server: Direct",
   "OpenEpic: eHealthExchange",
   "CernerHelios: eHealthExchange",
   "OPHDST Meld: Direct",
--- a/query-connector/src/app/database-service.ts
+++ b/query-connector/src/app/database-service.ts
@@ -192,7 +192,7 @@ export async function translateVSACToInternalValueSet(
   fhirValueset: FhirValueSet,
   ersdConceptType: ErsdConceptType,
 ) {
-  const id = fhirValueset.id;
+  const oid = fhirValueset.id;
   const version = fhirValueset.version;
 
   const name = fhirValueset.title;
@@ -205,10 +205,10 @@ export async function translateVSACToInternalValueSet(
   });
 
   return {
-    valueSetId: "",
+    valueSetId: `${oid}_${version}`,
     valueSetVersion: version,
     valueSetName: name,
-    valueSetExternalId: id,
+    valueSetExternalId: oid,
     author: author,
     system: system,
     ersdConceptType: ersdConceptType,
--- a/query-connector/src/app/database-service.ts
+++ b/query-connector/src/app/database-service.ts
@@ -10,9 +10,15 @@ import {
   ersdToDibbsConceptMap,
 } from "./constants";
 import { encode } from "base-64";
+import {
+  UserQueryInput,
+  generateUserQueryInsertionSql,
+  generateUserQueryToValueSetInsertionSql,
+} from "./query-building";
+import { UUID } from "crypto";
 
 const getQuerybyNameSQL = `
-select q.query_name, q.id, qtv.valueset_id, vs.name as valueset_name, vs.version, vs.author as author, vs.type, vs.dibbs_concept_type as dibbs_concept_type, qic.concept_id, qic.include, c.code, c.code_system, c.display 
+select q.query_name, q.id, qtv.valueset_id, vs.name as valueset_name, vs.oid as valueset_external_id, vs.version, vs.author as author, vs.type, vs.dibbs_concept_type as dibbs_concept_type, qic.concept_id, qic.include, c.code, c.code_system, c.display 
   from query q 
   left join query_to_valueset qtv on q.id = qtv.query_id 
   left join valuesets vs on qtv.valueset_id = vs.id
@@ -61,9 +67,7 @@ export const getSavedQueryByName = async (name: string) => {
  * @param rows The Rows returned from the DB Query.
  * @returns A list of ValueSets, which hold the Concepts pulled from the DB.
  */
-export const mapQueryRowsToConceptValueSets = async (
-  rows: QueryResultRow[],
-) => {
+export const mapQueryRowsToValueSets = async (rows: QueryResultRow[]) => {
   // Create groupings of rows (each of which is a single Concept) by their ValueSet ID
   const vsIdGroupedRows = rows.reduce((conceptsByVSId, r) => {
     if (!(r["valueset_id"] in conceptsByVSId)) {
@@ -81,6 +85,10 @@ export const mapQueryRowsToConceptValueSets = async (
       valueSetId: conceptGroup[0]["valueset_id"],
       valueSetVersion: conceptGroup[0]["version"],
       valueSetName: conceptGroup[0]["valueset_name"],
+      // External ID might not be defined for user-defined valuesets
+      valueSetExternalId: conceptGroup[0]["valueset_external_id"]
+        ? conceptGroup[0]["valueset_external_id"]
+        : undefined,
       author: conceptGroup[0]["author"],
       system: conceptGroup[0]["code_system"],
       ersdConceptType: conceptGroup[0]["type"],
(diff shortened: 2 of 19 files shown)


query-connector/src/app/database-service.ts
```diff
@@ -267,8 +276,11 @@ export async function insertValueSet(vs: ValueSet) {
...
 function generateValueSetSqlPromise(vs: ValueSet) {
-  const valueSetOid = vs.valueSetId;
+  const valueSetOid = vs.valueSetExternalId;
 
+  // TODO: based on how non-VSAC valuests are shaped in the future, we may need
+  // to update the ID scheme to have something more generically defined that
+  // don't rely on potentially null external ID values.
   const valueSetUniqueId = `${valueSetOid}_${vs.valueSetVersion}`;
   const insertValueSetSql =
     "INSERT INTO valuesets VALUES($1,$2,$3,$4,$5,$6) RETURNING id;";
```
m-goggins: I think I missed this before, but we don't expect users to provide an `ersdConceptType`, but they should provide a `DibbsConceptType` so we should be inserting a `DibbsConceptType`

query-connector/src/app/database-service.ts
```diff
@@ -391,3 +403,51 @@ function logRejectedPromiseReasons<T>(
...
+    queryId as UUID,
+  );
+
+  const joinPromises = insertJoinSqlArray.map((q) => {
+    dbClient.query(q.sql, q.values);
+  });
+
+  const joinInsertResults = await Promise.allSettled(joinPromises);
+
+  const jointInsertsSucceeded = joinInsertResults.every(
```
DanPaseltiner, line 442: Do we mean "join" or "joint" here?

query-connector/src/app/query-building.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Function that generates SQL for the query_to_valueset join table needed for
+ * query building.
+ * @param input - Values of the shape UserQueryInput needed for query insertion
+ * @param queryId - ID of the query that's already been created to associate with
+ * a given valueset
+ * @returns An array of {sql, values} to be inserted by the join insertion flow
+ */
+export function generateUserDefinedQueryToValueSetInsertionSql(
```
DanPaseltiner, line 54: Aren't all queries user defined?

query-connector/src/app/database-service.ts
```diff
@@ -391,3 +403,51 @@ function logRejectedPromiseReasons<T>(
       return r.reason;
     });
 }
+
+/**
+ * Function that orchestrates query insertion for the query building flow
+ * @param input - Values of the shape UserQueryInput needed for query insertion
+ * @returns - Success or failure status, with associated error message for frontend
+ */
+export async function insertUserDefinedQuery(input: UserQueryInput) {
```
DanPaseltiner, line 412: All queries will be user defined. I think we can just `insertQuery`.

m-goggins (commented):
The overall approach looks good to me. I left a couple of comments re: the eRSD vs. DIBBs concept types and what we expect users to be able to provide. I think it's reasonable for users to provide whether the valueset contains codes related to labs, medications, conditions, etc. (i.e., DIBBs concept type) but not the eRSD concept type.

DanPaseltiner (commented):
Looks pretty good to me. I think we can simplify some of our language here. The current plan is that we will only offer "query templates" out of the box and all queries available in the app will be user defined, therefore we don't need a distinction between user defined and non-user defined queries.

Also, how should we approach testing this work? Unless I'm missing something we don't have much by the way of tests written.

query-connector/src/app/constants.ts
```diff
@@ -319,6 +319,7 @@ export interface ValueSet {
   valueSetId: string;
   valueSetVersion: string;
   valueSetName: string;
+  valueSetExternalId?: string;
```
bamader, line 322: I thought we were punting on this schema change until Marcelle and I were caught up with existing DB changes and migrations?

query-connector/src/app/database-service.ts
```diff
@@ -81,6 +85,10 @@ export const mapQueryRowsToConceptValueSets = async (
       valueSetId: conceptGroup[0]["valueset_id"],
       valueSetVersion: conceptGroup[0]["version"],
       valueSetName: conceptGroup[0]["valueset_name"],
+      // External ID might not be defined for user-defined valuesets
+      valueSetExternalId: conceptGroup[0]["valueset_external_id"]
+        ? conceptGroup[0]["valueset_external_id"]
+        : undefined,
       author: conceptGroup[0]["author"],
       system: conceptGroup[0]["code_system"],
       ersdConceptType: conceptGroup[0]["type"],
```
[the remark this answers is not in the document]
bamader: ➕

query-connector/src/app/database-service.ts
bamader: ➕

query-connector/src/app/database-service.ts
```diff
@@ -391,3 +403,51 @@ function logRejectedPromiseReasons<T>(
...
     });
 }
+
+/**
+ * Function that orchestrates query insertion for the query building flow
+ * @param input - Values of the shape UserQueryInput needed for query insertion
+ * @returns - Success or failure status, with associated error message for frontend
+ */
+export async function insertUserDefinedQuery(input: UserQueryInput) {
+  const { sql, values } = generateUserDefinedQueryInsertionSql(input);
```
bamader: Similarly, I think this can just be `generateQueryInsertionSql`

query-connector/src/app/query-building.ts
bamader: Yeah I left this elsewhere but I'd probably say we can just make the nomenclature `generateQueryTo...` and drop the `UserDefined` across all instances

query-connector/src/app/database-service.ts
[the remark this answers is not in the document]
bamader: If we leave it in, that is

query-connector/src/app/database-service.ts
fzhao99: ahh good catch! changed

bamader (commented):
Approach seems good to me. I agree with Marcelle around the extent of what's reasonable to expect users to provide. The only thing I'm hung up on is the status of that `externalId` column on `ValueSet`. Based on standup, I thought we were going to hold off on changing the schema until the existing DB work was completed. I could be misunderstanding what was actually talked through, though.

query-connector/src/app/constants.ts
fzhao99: Ahh, apologies for not getting everyone on the same page. 

The conclusion I came away with was
- We agreed that it's helpful / not too disruptive to include the VSAC OID in schema for use elsewhere. Am reasoning that since we're adding a new property, things won't be too disruptive, but let me know if you feel differently.
- Instead of naming it `valueSetOid`, we can name it `valueSetExternalId` since it generically covers future non-VSAC ID cases. This is mostly a cosmetic change since the underlying ID itself is always going to be OID now until we support a non-VSAC use case. That is, we're not changing anything about the underlying schema other than including a newly named field that used to be there.
- Since we don't know the future shape / conventions of ID's from other terminologies, we're going to punt on how we replace the actual values that we use to construct the ID's. That is, we might change the way we construct `valueSetId`'s in the future to potentially not use this `externalId` value. This is a non-cosmetic change to how external ID's interact with the primary keys for the table. 

Hopefully this clarifies things, but if not, happy to hop on a call to explain further. Happy also just to change this to `valueSetOid` if that makes things more straightforawrd: not too attached to the naming here as long as we can have the field broken out for use elsewhere.

fzhao99:
> Looks pretty good to me. I think we can simplify some of our language here. The current plan is that we will only offer "query templates" out of the box and all queries available in the app will be user defined, therefore we don't need a distinction between user defined and non-user defined queries.
> 
> Also, how should we approach testing this work? Unless I'm missing something we don't have much by the way of tests written.

Good question! Ideally we'd have an integration test that does a straightforward insert / validates the database state that it indeed inserted the query correctly, especially with the cascading joins. Since we don't have the integration test setup currently though, I elected the manual route with the test branch instead to show the simple case that the happy path works. Once we get the integration tests working again, happy to write a test that validate the query gets created correctly when fed the right frontend input.

In the meantime, once we get enough of the frontend built to warrant an e2e, would also be happy to write one of those to cover a more interesting chunk of the feature.

query-connector/src/app/constants.ts
bamader: Ahhh that third bullet point helps clarify things a lot. Thanks for the write up here, Bob. This definitely clarifies the intent and I agree as it stands this isn't particularly disruptive or blocking. Just wanted to get back on the same page, but I'm there now. Thanks!

bamader (approved):
Everything looks good now that I'm back up to speed. I also agree with Bob's approach to testing at the moment--a manual route with an eventual e2e once more of the front end is built out seems like the proper way to do things. An integration or other intermediary test will just be outmoded as soon as we have a front end to test with fully imo.

m-goggins (approved):
Thanks for making those changes!

robertandremitchell (approved):
looks good!

query-connector/src/app/query-building.ts
```diff
@@ -0,0 +1,72 @@
+import { UUID, randomUUID } from "crypto";
+import { Concept, ValueSet } from "./constants";
+
+// TODO: Potentially merge this / infer this from the type created via the
+// database creation workstream
+export type UserQueryInput = {
+  queryName: string;
+  author: string;
+  valueSets: ValueSet[];
+  concepts?: Concept[];
```
m-goggins, line 10: Sorry one more question: why do we need `concepts` here if concepts are included in the `ValueSet` type?

fzhao99: I included this to allow for additional insertions to the `query_included_concepts` table in the future where we might have concepts without parent valuesets. Probably won't use this for a while until users can add their own codes, but just included it for completeness since we have that table already

m-goggins: Missed this comment earlier. We shouldn't have concepts without valuesets. If a user wants to create a concept, e.g., add a single code, they would still need to make a valueset for that single concept.

fzhao99: Ahh ok! Will go back and remove this then. Glad we caught it

PR #90, opened 2024-10-29
Set up local HAPI server for e2e tests

m-goggins:
# PULL REQUEST

## Summary

This PR adds a new `docker-compose-e2e-tests.yaml` that spins up a local HAPI FHIR server and loads the Hyper Unlucky patient into the server. I chose to add a new dock compose file to prevent the primary, `docker compose` file from requiring HAPI, which is only used for e2e tests, but open to hearing others' ideas.

I also updated the e2e tests to manually select the local instance of the HAPI server, which is now available in the drop down. 

Some lessons learned:

- Our flaky issues stemmed from the Public HAPI server, which we thought was the issue, but it was nice to get additional confirmation. We should probably continue to make sure our e2e tests remain divorced from external services.
- We did not end up needing to use networks or the container names when connecting between containers either locally or in github actions. The only place where the container name is used instead of localhost is in `fhir_servers`. 

## Related Issue

Fixes # [phdi 2758](https://app.zenhub.com/workspaces/skylight-dibbs-viper-6480ba23b5a07644e0e46d23/issues/gh/cdcgov/phdi/2758)

## Acceptance Criteria

## Checklist

### DIFF
--- a/query-connector/src/app/constants.ts
+++ b/query-connector/src/app/constants.ts
@@ -387,3 +387,18 @@ export function isFhirResource(resource: unknown): resource is FhirResource {
     "resourceType" in resource
   );
 }
+
+// The value for "concept version" (sometimes) exists under "expansion" in the VSAC FHIR
+// response. This is similar, but different in some subtle ways from the "compose.include"
+// path that we're currently using to grab concept information. Although we could
+// grab and parse this information from the FHIR response, it would involve
+// changing our data model to store information that we think is a "nice to
+// have", which is only available sometimes. As a result, we're purposefully
+// leaving this blank until we can clean up the migration schema to drop these columns
+export const INTENTIONAL_EMPTY_STRING_FOR_CONCEPT_VERSION = "";
+
+// Originally, the column in the concept table was set up to maintain backwards e
+//compatibility with the ICD-9 codes that the team is deciding not to support after
+// we clean up the DB migration. Leaving these in until we can clean these up
+// in the migration schema
+export const INTENTIONAL_EMPTY_STRING_FOR_GEM_CODE = "";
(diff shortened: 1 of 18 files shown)


query-connector/src/app/constants.ts
```diff
@@ -81,6 +81,7 @@ export const FhirServers = [
   "JMC Meld: Direct",
   "JMC Meld: eHealthExchange",
   "Public HAPI: Direct",
+  "Local e2e HAPI Server: Direct",
```
DanPaseltiner, line 84: Are we now going to display this as an option in the FHIR server drop down in the UI? I think we probably want to avoid that

DanPaseltiner (approved):
LGTM. Thanks for sorting all of this out!

bamader (approved):
I am so happy this is done 😭 😎

query-connector/src/app/constants.ts
fzhao99: Could maybe do a check for whether this is at the live domain and filter it out if it is

fzhao99 (approved):
thanks so much for doing this! 🤞 this solves all our problems

query-connector/src/app/constants.ts
m-goggins: Yeah, it's currently being displayed. I couldn't figure out a way to make it work another way on the e2e tests besides selecting the server from the drop down. I figured that it was behind an "Advanced" label and I haven't seen folks attempting to query other servers, it would be okay. 

My inclination would be to change the FHIR server option within the tests. I'm not sure how that would be done exactly, but I can look to see if it's possible in playwright.

DanPaseltiner: Ehh maybe this isn't worth the hassle atm

PR #91, opened 2024-10-29
Allow overwriting inserts

### LINKED ISSUE #81: Change ValueSet insertion function to overwrite rather than error if duplicate entries are inserted
Continuing with the work outlined in [this document](https://docs.google.com/document/d/10Aw7OKOD4Bw7zWAbsqYS-Eu8mGg0AGjc_5auow0DbQA/edit), we want our inserts to be idempotent and failure resilient, so that any errors that occur during database creation don't break the process completely. One minor change that will be needed to accomplish this is making the `insertValueSet` function use an `UPSERT` rather than a true `INSERT`, so that if duplicate values are created during valueset initialization the DB migration can continue. Follow the syntax [here](https://neon.tech/postgresql/postgresql-tutorial/postgresql-upsert) to modify this function to use an `INSERT … ON CONFLICT` pattern to resolve any duplications in the uniquely identifying OID + Version number field.

bamader:
# Allow overwriting inserts

## Summary

This PR modifies the `INSERT` statement of our `InsertValueSet` function to allow for duplicated uniqueIDs to overwrite those of existing entries. Our DB creation process will go through a fair number of duplicate value sets after pulling down the eRSD, so simply overwriting them with the same data will allow the seeding process to proceed more smoothly than error-ing and breaking the transaction.

## Related Issue

Fixes #81 

## Additional Information

I created a test branch here https://github.com/CDCgov/dibbs-query-connector/tree/test-vs-overwrite that removes database seeding in migrations 2 through 6, and instead adds a button to the landing page that tests a couple cases of inserting valuesets and duplicates into the DB. No errors get thrown and viewing the DB in DBeaver shows all rows correctly entered.

### DIFF
--- a/query-connector/src/app/database-service.ts
+++ b/query-connector/src/app/database-service.ts
@@ -270,8 +270,25 @@ function generateValueSetSqlPromise(vs: ValueSet) {
   const valueSetOid = vs.valueSetId;
 
   const valueSetUniqueId = `${valueSetOid}_${vs.valueSetVersion}`;
-  const insertValueSetSql =
-    "INSERT INTO valuesets VALUES($1,$2,$3,$4,$5,$6) RETURNING id;";
+
+  // In the event a duplicate value set by OID + Version is entered, simply
+  // update the existing one to have the new set of information
+  // ValueSets are already uniquely identified by OID + V so this just allows
+  // us to proceed with DB creation in the event a duplicate VS from another
+  // group is pulled and loaded
+  const insertValueSetSql = `
+  INSERT INTO valuesets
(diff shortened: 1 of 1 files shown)


bamader:
Separate testing branch (like Bob did when first working on the insert value set) is still underway

m-goggins (approved):
LGTM! Just want to get @DanPaseltiner to take a look to make sure we correctly understood what he wanted here.

bamader:
Adding Dan as a reviewer to get any thoughts

DanPaseltiner (approved):
🚀

PR #101, opened 2024-10-31
backend query insertion cleanup

fzhao99:
Followup to #89 as raised by Marcelle [here](https://github.com/CDCgov/dibbs-query-connector/pull/89#discussion_r1824443118)

Also renaming some things to remove an extra noun as discussed in the other PR

Also ~fixes~ improves the e2e's by changing the way we set the fetched valueSet data in the customize query page


query-connector/src/app/query/components/CustomizeQuery.tsx
```diff
@@ -46,41 +51,35 @@ const CustomizeQuery: React.FC<CustomizeQueryProps> = ({
   goBack,
 }) => {
   const [activeTab, setActiveTab] = useState<DibbsValueSetType>("labs");
-  const { labs, conditions, medications } =
-    mapValueSetsToValueSetTypes(queryValuesets);
-  const [valueSetOptions, setValueSetOptions] = useState({
```
fzhao99, line 51: Setting the state this way was causing the tests to not re-render the component when the value for `queryValuesets` updated. As a result, even when the data came back from the backend, the page was rendering as if nothing was updating. 

Rewriting this to use a `useEffect` instead guarentees that React with reset the state correctly when the value of `queryValuesets` updates

bamader: This makes perfect sense and is a prime example of why, not a front-end person, know _just_ enough react to be an absolute hailstorm of danger to all proper stateful patterns 😅

bamader (approved):
Great catch!

Issue #73, opened 2024-10-28
Set up routing to get to query building


fzhao99:
* Path to the beginning of the flow from [the gear icon → manage custom queries](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=480-33547&t=nCvdKQwvRKVyZnv4-4)

  <img src="https://lh7-rt.googleusercontent.com/docsz/AD_4nXfasC6YB4QEAfiwMKZfC9EvH-TuMpHyyI4vqbNINNdmQL4Z5cEtUTW_9InShuRGOPOubuJFwKjwp6voEnRp_GRmws3cgB0Tu8U7VP28-SkfTgYg19_AE5bhQqFFU4u8wAZGBJZ1QmzrWhb_5ezoRYEL2Lg1?key=PGIkf2d1l2H6joAt59FI5JtV " alt="" width="624" height="360" />
* Will implement this using the "orphan page strategy" that we used for the customize query flow previously 
  * This will entail us not linking to the pages in the UI but making it available through a direct URL that we can use for review. This will allow us to not have to keep a long running feature branch / load in smaller units of code.
  * Depending on the timing of the CD work, we might need to coordinate around how “accessible” this is to our users.

PR #107, opened 2024-10-31
setup routing and empty state query building pages

### LINKED ISSUE #74: Implement empty customize query page state
[Empty state designs](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=480-33625&t=nCvdKQwvRKVyZnv4-4)

fzhao99:
# PULL REQUEST

## Summary
First piece of the query building frontend work that sets up the empty state page and routing. Turning off the navigational link in production, but accessible locally via the gear icon once signed in / in demo. Can also navigate to the empty state page at `/queryBuilding`

### For design: 
Once you navigate to the query building page, do a generic sweep to make sure the empty state page matches designs - spacing / sizing / color of the empty state box pages

Did also tokenize the font styles for h1/h2's (to Merriweather), body (to Public Sans Web), and the site title / mono (to Consolas), so do a step-through of the pages in the rest of the app to ensure those fonts are applied correctly.

## Related Issue
Fixes #74 and #73 

## Additional Information
- Tokenized the fonts of the header files to use the new text styles that we're using across the application using the USWDS stylings. Updated heading, body, site title to match
- Tokenized some colors for use in a few places with the changes to the spacing
- Moved the `SelectQuery.tsx` file out of the root `query` directory
- Moved the `stepIndicator` folder into the `components directory
- Collapsed two utils files into one
- Deleted some unnecessary stylings

## Checklist

- [x] Link to relevant issues
- [x] Update documentation


query-connector/src/app/query/components/customizeQuery/customizeQueryUtils.ts
```diff
@@ -138,3 +138,20 @@ export const mapValueSetsToValueSetType = (valueSets: ValueSet[]) => {
 
   return results;
 };
+
```
fzhao99, line 141: moved this from an orphan utils file into this one since we had it already

robertandremitchell (commented):
looks good! just to note, for docker, when I tried to follow the logic to sign-in, my little dropdown only had the sign out; I could only get the `My queries` button on `npm run dev` so we might need another way for @mikang to test since that workflow uses Docker. Or commenting out !isProduction until we merge so that docker can be used to test the workflow.

fzhao99:
> I could only get the My queries button on npm run dev so we might need another way for @mikang to test since that workflow uses Docker.

Ahh yes good point! Realized the docker setups gauge those production flags differently. 

The page is accessible via URL at /queryBuilding so we can do that too.

katyasoup (commented):
looks great! nice work on the cleanup bits, too 😃

query-connector/src/app/layout.tsx
```diff
@@ -1,6 +1,6 @@
 import "../styles/styles.scss";
 import Header from "./query/components/header/header";
-import Footer from "./footer";
+import Footer from "./query/components/footer/footer";
```
DanPaseltiner, line 3: Not blocking for this PR, but it seems like an organizational anti-pattern here to have `components` nested within `query` if these components are going to be used in other parts of the app. I don't want to increase our current scope here, but would like to flag this for the team so we can discuss further.

query-connector/src/app/page.tsx
```diff
@@ -7,15 +7,15 @@ import {
...
-
+import styles from "./page.module.scss";
 /**
  * The landing page for the TEFCA Viewer.
  * @returns The LandingPage component.
  */
 export default function LandingPage() {
   const router = useRouter();
 
-  const handleClick = () => {
```
DanPaseltiner, line 18: Really appreciate this. Thanks!

DanPaseltiner (approved):
LGTM 🚀 . Couple comments below but nothing is blocking.

query-connector/src/app/layout.tsx
fzhao99: Yep, totally makes sense. Will bring it up in a shared forum!

Issue #84, opened 2024-10-29
Write counting function to verify value set insertion during eRSD pull


bamader:
We've opted to pursue a strategy of not wrapping the eRSD pull and value set insertions in a transaction, since that would either be a) slow and painful at the concept/value set level, or b) monolithically unwieldy if the entire process were wrapped in a transaction. Instead, what we'll do for each value set we insert is to use a "sanity check" function to make sure that everything got inserted. Given a particular value set, this function will query the eRSD (or parse our pulled extract of it, either way works so long as it's the complete source-of-truth eRSD data) to determine how many concepts / pieces of data should have been inserted into the DB as part of adding this value set. It will also identify which concepts / pieces of data those are, e.g. by name or ID. With this list in hand, the function will query our database after the insert call has been performed to see how many concepts come back. If these numbers agree, and if the concepts are all the same in both cases, we consider the insertion a success and move on to the next valueset. If the numbers disagree or the data don't match, we flag the value set in question for a repeat pull / insert at the end of the DB creation process, reporting to the user that there were problems with it as we try to fix them. This repeat process is why we want the insert to replace entries on conflict, so that a value set that errors doesn't break this process by having multiple copies of the same concept after we attempt to pull it again.

PR #108, opened 2024-11-01
Vs insertion check

bamader:
# Sanity Check Value Set Insertion

## Summary

This PR introduces a function that verifies that a value set has been successfully inserted into our Postgres DB. Given a FHIR-formatted value set, the function checks that the value set itself has been inserted, all related concepts have been inserted, and all foreign key associations between the two are stored in the relevant tables. This PR also corrects a bug that was introduced when trying to add external IDs, namely that ValueSet IDs were having the version appended twice, which was causing cascading failures when inserting.

## Related Issue

Fixes https://github.com/CDCgov/dibbs-query-connector/issues/84

## Additional Information

As usual, I've built a test branch with a button that allows you to check the integrity of the insertion and sanity counting code. It's accessible here: https://github.com/CDCgov/dibbs-query-connector/tree/test-vs-count . Start the app, open the console, then click the button to see some tests happen.

### DIFF
--- a/query-connector/src/app/database-service.ts
+++ b/query-connector/src/app/database-service.ts
@@ -461,17 +461,12 @@ export async function insertQuery(input: QueryInput) {
  *   2. Whether each concept included in that value set bundle was inserted
  *   3. Whether these concepts are now mapped to this value set via foreign key
  * If any data is found to be missing, it is collected and logged to the user.
- * @param setToCheck The FHIR formatted value set to check.
- * @param ersdConcept The eRSD concept this value set belongs to.
+ * @param vs The DIBBs internal representation of the value set to check.
  * @returns A data structure reporting on missing concepts or value set links.
  */
-export async function checkValueSetInsertion(
-  setToCheck: FhirValueSet,
-  ersdConcept: ErsdConceptType,
-) {
+export async function checkValueSetInsertion(vs: ValueSet) {
   // Translate to our internal representation and begin accumulating
   // missing data
-  const vs = await translateVSACToInternalValueSet(setToCheck, ersdConcept);
   const missingData = {
     missingValueSet: false,
     missingConcepts: [] as Array<String>,
--- a/query-connector/src/app/database-service.ts
+++ b/query-connector/src/app/database-service.ts
@@ -465,8 +465,7 @@ export async function insertQuery(input: QueryInput) {
  * @returns A data structure reporting on missing concepts or value set links.
  */
 export async function checkValueSetInsertion(vs: ValueSet) {
-  // Translate to our internal representation and begin accumulating
-  // missing data
+  // Begin accumulating missing data
   const missingData = {
     missingValueSet: false,
     missingConcepts: [] as Array<String>,
--- a/query-connector/src/app/database-service.ts
+++ b/query-connector/src/app/database-service.ts
@@ -358,7 +358,6 @@ function generateConceptSqlPromises(vs: ValueSet) {
 }
 
 function generateValuesetConceptJoinSqlPromises(vs: ValueSet) {
-  const valueSetUniqueId = `${vs.valueSetId}_${vs.valueSetVersion}`;
   const insertConceptsSqlArray = vs.concepts.map((concept) => {
     const systemPrefix = stripProtocolAndTLDFromSystemUrl(vs.system);
     const conceptUniqueId = `${systemPrefix}_${concept.code}`;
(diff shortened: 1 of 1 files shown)


query-connector/src/app/database-service.ts
```diff
@@ -377,8 +376,8 @@ function generateValuesetConceptJoinSqlPromises(vs: ValueSet) {
     RETURNING valueset_id, concept_id;
     `;
     const conceptInsertPromise = dbClient.query(insertJoinSql, [
-      `${valueSetUniqueId}_${conceptUniqueId}`,
-      valueSetUniqueId,
+      `${vs.valueSetId}_${conceptUniqueId}`,
+      vs.valueSetId,
```
m-goggins, line 380: nice catch!

query-connector/src/app/database-service.ts
```diff
@@ -453,3 +452,118 @@ export async function insertQuery(input: QueryInput) {
...
+      error,
+    );
+    const systemPrefix = stripProtocolAndTLDFromSystemUrl(vs.system);
+    vs.concepts.forEach((c) => {
+      const conceptUniqueId = `${systemPrefix}_${c.code}`;
+      missingData.missingMappings.push(conceptUniqueId);
+    });
+  }
+
+  return missingData;
```
m-goggins, line 568: Are you thinking that we might display this back to the user somehow?

query-connector/src/app/database-service.ts
```diff
@@ -453,3 +452,118 @@ export async function insertQuery(input: QueryInput) {
...
+ *   1. Whether the value set itself was inserted
+ *   2. Whether each concept included in that value set bundle was inserted
+ *   3. Whether these concepts are now mapped to this value set via foreign key
+ * If any data is found to be missing, it is collected and logged to the user.
+ * @param setToCheck The FHIR formatted value set to check.
+ * @param ersdConcept The eRSD concept this value set belongs to.
+ * @returns A data structure reporting on missing concepts or value set links.
+ */
+export async function checkValueSetInsertion(
+  setToCheck: FhirValueSet,
```
m-goggins, line 469: Should this be a FhirValueSet or the internal one? For any valuesets that are not from the eRSD, e.g., ones that are manually entered they will be in the internal valueset format (I believe). To make this function extensible to all the valuesets we expect, I think it should expect an internal model of a valueset. 

My understand is that we go: whatever format -> internal vs -> insert so we should always check against the internal model.

m-goggins (changes requested):
Looks good overall, but I think the check function should have the internal valueset model as the input, not FHIR, so that it can apply to all valuesets that we may insert not just the ones from the eRSD.

query-connector/src/app/database-service.ts
bamader: This thing gave me no end of problems when debugging. Not even the regular insert was working because the valueset got inserted regularly but then the foreign keys for `valueset_to_concept` all had the version number appended a second time so the DB was imploding 😭

query-connector/src/app/database-service.ts
bamader: This makes sense to me. It doesn't impact any of the logic otherwise and keeps us more robust to different input sources so I like this improvement.

query-connector/src/app/database-service.ts
bamader: I think it could have a lot of different uses. I think display is one of them, but I was actually just imagining it as a data structure for the insert script to use. There are different cases of missingness if an insertion fails, and we need to be able to capture the extent to which things are missing. E.g., we might have successfully inserted all concepts, but not the value set itself, which means we also failed on foreign key mappings. I thought a succinct object that captured each of these dimensions separately was the cleanest way for another part of the code (which I think is a separate helper function when we write the script, something like

```
handleValueSetReInsertion(missingData) {
  if (missingData.missingValueSet) { ... }
  ...
}
```
) to then go back in and take care of whatever we were missing in a graceful and organized way. I imagined this as fairly internal which is why I didn't make an explicit type for it, but I can if we intend to display it.

m-goggins: Gotcha! That makes sense to me.

m-goggins (approved):
Looks good! Thanks for making the changes to the internal VS

query-connector/src/app/database-service.ts
```diff
@@ -453,3 +452,113 @@ export async function insertQuery(input: QueryInput) {
...
+ * value set, the function checks three things:
+ *   1. Whether the value set itself was inserted
+ *   2. Whether each concept included in that value set bundle was inserted
+ *   3. Whether these concepts are now mapped to this value set via foreign key
+ * If any data is found to be missing, it is collected and logged to the user.
+ * @param vs The DIBBs internal representation of the value set to check.
+ * @returns A data structure reporting on missing concepts or value set links.
+ */
+export async function checkValueSetInsertion(vs: ValueSet) {
+  // Translate to our internal representation and begin accumulating
```
m-goggins, line 468:
```suggestion
  // Begin accumulating
```

PR #111, opened 2024-11-04
fix sign in page height

fzhao99:
sets the sign in page content to fill up the whole width between the header / footer since I pulled out the grid styling on the main container that made it work before 😅 .

### DIFF
--- a/query-connector/src/app/query/components/footer/footer.module.scss
+++ b/query-connector/src/app/query/components/footer/footer.module.scss
@@ -1,7 +1,7 @@
 @use "../../../../styles/_variables" as *;
 
 .footerContainer {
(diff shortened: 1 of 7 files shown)


mikang:
Thank you @fzhao99 ! 

There's one more thing I just noticed - now it looks like the footer doesn't go full width / there's some whitespace between the footer and the edge of the browser @fzhao99 @katyasoup 

image

fzhao99:
To close the loop: Michelle and I chatted about this and for some reason, the gap is only showing up on her computer. It's minor enough, esp if Katie it looks fine to you, that we can merge this and debug it later!

katyasoup (approved):
🎉

PR #112, opened 2024-11-04
Updating condition category mapping work in database-services

robertandremitchell:
# PULL REQUEST

## Summary

Per conversation in 11/4 call, we do not want to use an API call for conditions data. To do this instead, I have opted ot try to use one call in the database-services.ts to ensure we have the main JSONs we should need: 
1. a JSON that is grouped by category for ease of displaying the name of each condition. The ID is included so that in the search function that will be on the page, someone could also search by ID.
2. a JSON of id and name. This is done because in order to make selections, we will need the IDs to look up and pull the valueset/concepts for display on the next page.

We could plausibly do this one with just the first JSON if we want to have additional transformation work, but because each of these should be negligible in size, it seems better to just do the work here?

## Related Issue

Fixes https://linear.app/skylight-cdc/issue/QUE-26/expose-category-information-to-the-frontend.

## Checklist

- [ X ] Descriptive Pull Request title
- [ X ] Link to relevant issues
- [ N/A ] Provide necessary context for design reviewers
- [ N/A ] Update documentation


fzhao99 (approved):
🚢

Issue #85, opened 2024-10-29
Write DB Creation script


bamader:
With all other pieces of functionality in place, we can finally write a script to create the DB once and for all. This bash script (or python, or whatever other DB creation language the dev who picks this up chooses) should be run after the flyway table creation migration is performed. It should pull down the whole eRSD as a FHIR bundle, then iterate through it and perform any and all necessary insertions. When a value set or group of data is finished, the sanity check function should be called to verify that all data was added, and any pieces that weren't properly inserted should be repeated at the end of the script.

PR #130, opened 2024-11-07
DB Creation Funcs

### LINKED ISSUE #83: Update DB Migrations
Once we have value set insertion and default queries handled, we'll need to update the flyway migration files to be responsible only for table creation. This means that we'll grab the column additions and addendums from some of the later migrations and squash them back into migration 1, when the tables are created. We'll also get rid of the migrations that make manual data insertions 100,000 times, deleting them entirely. The end result is that we should have a single migrations file that creates all tables in the final state we want them to appear.

bamader:
# Create DB Seeding Functions in TypeScript file

## Summary

This PR adds in the functionality needed to execute database creation as a currently uncalled typescript file. Once Marcelle's PR for creating the button and page lands, we can link this code to that button and play it out. In the mean time, this code was tested using docker instrumentation hook setups.

## Related Issue

Fixes #83 #85 

## Additional Information

Very minor point, but I personally like the logging I left in the functions because this is a more complex integration process across multiple APIs, with batching and async calls, as well as sleep times. I personally think it's nice to report the information since the whole process takes ~30 seconds and visual cues can be helpful during that time. Open to taking them out, just wanted to explain the rationale for leaving them in.


robertandremitchell (approved):
makes sense to me!

PR #161, opened 2024-11-14
Save condition selections on template builder page

### LINKED ISSUE #118: Update the read-only query template selection page to do a write / save
## Description:

Take the "read only" condition template page and ~~make it write~~ save the data for use in the next step(s)

## Acceptance Criteria:

* Once users select "save" on the query template building page, a query with their selected templates ~~gets saved to the database.~~ is saved in state
* Display interactions for the query name form field / create query button work as is discussed in [this Slack thread](https://skylight-hq.slack.com/archives/C07PQ078KK2/p1730997768181319)

katyasoup:
# PULL REQUEST

## Summary
- Update condition selection page to:
  - Autofocus the Query Name input on page load/refresh (if the input is blank)
  - Indicate that Query Name is required
- Add `selectedConditions` to component's state
  - Toggling the checkbox will now add/remove the given condition from this group 

## Related Issue

Fixes #118 

## Additional Information
- Original AC was to save data from this page to the db; since we are/were not saving query-condition relations, this isn't appropriate. Instead, the data is saved in the component state and will be used to retrieve the corresponding value sets on the next page (eventually sent to the db with the full query data (value sets, concepts, etc)).

## How to Test
- Navigate to the "My Queries" page: `landing page` -> `gear icon` -> `sign in` -> `gear icon` -> `my queries` -> `create query button`
  - The "Query Name" input should autofocus and the "Create Query" button should be disabled
  -  On hover over the button, tooltip text should read "Select at least one condition below"
- Select one or more conditions from at least two categories
  - The "Create Query" button should now be enabled
- Click to create the query, without adding a query name. 
  - An inline error should appear and focus should jump back to the Query Name input 
  - Tooltip text should read "Enter a query name, then click to create your query"
- Name your query
  - The inline error should go away
  - Tooltip text should read "Click to create your query"
- Open the dev console
- Click "Create Query" and confirm all your selections are logged appropriately
- Uncheck one condition and repeat above
  - Confirm the unchecked item was removed
  - Repeat such that a given category no longer has any checked items; confirm the category is no longer logged in the console 

Ex:
Screenshot 2024-11-14 at 17 19 46
Screenshot 2024-11-14 at 17 20 03

## Checklist

- [x] Link to relevant issues
- [ ] Update documentation


query-connector/src/app/queryBuilding/buildFromTemplates/ConditionColumnDisplay.tsx
```diff
@@ -56,9 +66,36 @@ export const ConditionColumnDisplay: React.FC<ConditionColumnDisplayProps> = ({
...
       include: !prevValues.include,
     };
+    const shouldRemove = prevFetch[category][conditionId].include == false;
+    updateSelectedConditions(shouldRemove, category, conditionId, prevFetch);
     setFetchedConditions(prevFetch);
   }
 
+  const updateSelectedConditions = (
+    shouldRemove: boolean,
+    category: string,
```
robertandremitchell, line 76: do we want to preserve at the category level? or even have a JSON versus an array?

I think we would rather preserve just at the `conditionId` level as the keys since we don't actually need categories for the next page or the name of the condition, but just the `conditionId` to do a look up to display. We could plausibly do that with just an array and pop/append as needed.

I don't think there's any harm in how it is now though; I think it could even be helpful if we want to ensure on the next page the conditions are "grouped" by category, even if it's not displayed.

query-connector/src/app/queryBuilding/buildFromTemplates/page.tsx
```diff
@@ -88,7 +106,8 @@ export default function QueryTemplateSelection() {
             <Button
               className="margin-0"
               type={"button"}
-              disabled={!noTemplateSelected}
+              disabled={!atLeastOneItemSelected || !queryName}
```
robertandremitchell, line 109: in the figma, it looks like the button is completely hidden before a selection was made. I think the disabled state better fits what we have on other pages, but i also know you raised on prior tickets that the disabled state is not generally preferred to just providing an error message indicating what they need to include. Not sure it needs to be fixed here though 🤷🏽

robertandremitchell (commented):
looks good! a few considerations but I don't think any rising to the level of a blocker

query-connector/src/app/queryBuilding/buildFromTemplates/ConditionColumnDisplay.tsx
katyasoup: yeah, part of my thinking here was knowing we are tweaking how all the query data gets stored in the db - so wanting to preserve all the info we have now and we can always ease up if/when we decide we don't need it to be either stored in the db or shown in the UI. happy to rework later on if we decide it's too complicated/unnecessary here!

query-connector/src/app/queryBuilding/buildFromTemplates/page.tsx
katyasoup: ah, good catch! I will rework this per Michelle's revisions (from this [Slack thread](https://skylight-hq.slack.com/archives/C07PQ078KK2/p1731077429943419?thread_ts=1730997768.181319&cid=C07PQ078KK2)) , so that the button is hidden in an empty state, but shown and clickable when conditions are selected (which would trigger an error message if there is no Query Name input yet).

katyasoup:
Screenshot 2024-11-19 at 18 10 28
Screenshot 2024-11-19 at 18 10 38
Screenshot 2024-11-19 at 18 11 50
Screenshot 2024-11-19 at 18 10 53
Screenshot 2024-11-19 at 18 11 02

@mikang @robertandremitchell button state and inline error updated as shown above (I can't capture the cursor in a screenshot but it shows either a pointer or an X based on disabled status).

robertandremitchell (approved):
looks good to me!

mikang:
Beautiful! Thank you @katyasoup !

PR #166, opened 2024-11-19
Replace non-numeric characters in phone numbers with url encoding

### LINKED ISSUE #167: Replace non-numeric characters in phone numbers with url encoding
## Description:

Found a bug in the way that we generate the list of phone numbers in our queries while working on [QUE-52](https://linear.app/skylight-cdc/issue/QUE-52/update-postman-examples). This ticket fixes the bug by changing the way we generate the list of phone numbers to search for by replacing non-numeric characters, e.g. `+` and `-`, with their url encoded equivalents. This allows us to query for multiple phone numbers via eHealth Exchange, which eHealth devs recommend. 

## Acceptance Criteria:

When a user make a query that includes a phone number either through the API or UI, the generated list of phone numbers uses url encoding and the query is successful.

m-goggins:
# PULL REQUEST

## Summary

This PR changes the way that we generate the list of phone numbers to search for by replacing non-numeric characters, e.g. `+` and `-`, with their url encoded equivalents. This allows us to query for multiple phone numbers via eHealth Exchange, which eHealth devs recommend. 

## Related Issue

Fixes #167 

## Additional Information

We may look into more robust URL encoding for other fields, e.g., first names with spaces.

## Checklist

- [X] Descriptive Pull Request title
- [X] Link to relevant issues
- [ ] Update documentation

### DIFF
--- a/query-connector/src/app/format-service.tsx
+++ b/query-connector/src/app/format-service.tsx
@@ -269,7 +269,7 @@ export async function GetPhoneQueryFormats(phone: string) {
  */
 export const formatValueSetsAsQuerySpec = async (
   useCase: string,
-  valueSets: ValueSet[]
+  valueSets: ValueSet[],
 ) => {
   let secondEncounter: boolean = false;
   if (["cancer", "chlamydia", "gonorrhea", "syphilis"].includes(useCase)) {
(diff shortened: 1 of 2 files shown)


robertandremitchell (approved):
lgtm, i wonder if we want to use something out of the box like  encodeURI https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI

Something like that will probably be needed for names, probably less important for phone numbers which have a very limited number of special characters.

m-goggins:
Nice find @robertandremitchell! I added your suggestion (had to use encodeURIComponent) to get it to work on the full list. I'm going to follow up this ticket with a more comprehensive look at how we can do encoding for all input fields since I know first name also can have this issue if you have two given names with a space, e.g., 'Jacob Jingleheimer`

PR #174, opened 2024-11-25
Changing conflicting ports in docker-compose-dev

robertandremitchell:
# PULL REQUEST

## Summary

There's two `8080:8080` ports that are causing `npm run dev` to fail due to the bind already existing.

## Related Issue

Fixes baby bug.

## Additional Information

Anything else the review team should know?

## Checklist

- [x] Descriptive Pull Request title
- [ ] Link to relevant issues
- [x] Provide necessary context for design reviewers
- [x] Update documentation

PR #175, opened 2024-11-25
Make sure keycloak port bind is bound to 8080 inside container

nickclyde:
Follow up to #174 - the internal port bind needs to be 8080 since keycloak runs on 8080 by default inside the container. Also updated the `.env.sample` to reflect the new port in the issuer URL.


katyasoup (approved):
🙏🏻

robertandremitchell (approved):
oops, thanks for catching!

PR #178, opened 2024-11-26
use flex to size footer, header, and body

fzhao99:
# PULL REQUEST

## Summary
The designers were noticing that there was extra padding in the header / footer on some browsers. Had trouble reproducing, but think the issue was some 100vh CSS that was causing the spacing fill to behave weirdly.

image

## Related Issue

Fixes #

## Additional Information

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation

### DIFF
--- a/query-connector/src/styles/custom-styles.scss
+++ b/query-connector/src/styles/custom-styles.scss
@@ -1,17 +1,25 @@
 @use "uswds-core" as *;
 @use "_variables" as *;
 
+html,
+body {
+  height: 100%;
+}
 html {
   scroll-behavior: auto;
+  box-sizing: border-box;
 }
 
 body {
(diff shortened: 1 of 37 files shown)


query-connector/src/styles/custom-styles.scss
```diff
@@ -26,7 +34,7 @@ h1 {
 
 .main-body {
   background-color: rgba(247, 249, 250, 1);
-  flex: auto; // Sizes main-body to fill space between header and footer
+  flex: 1 1 auto;
```
katyasoup, line 37: could we add this to `main-container` as well, so other pages can have the footer anchored to the bottom?

katyasoup (approved):
lgtm! one small suggestion if its easy to squeeze in

PR #202, opened 2024-12-06
Add Copy ID / Delete functionality to My Queries page

### LINKED ISSUE #136: Edit / delete / create query - actions exist that route you to the condition template page
## Description:

On the my query page, the icons for deletion works / editing take you to the appropriate place on [the condition template selection page](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=475-20666&m=dev)

## Acceptance Criteria:

* On clicking the deletion icon, query deletes as expected
* On clicking the editing icon, the condition template selection page renders as expected

robertandremitchell:
# PULL REQUEST

## Summary

Copy ID
* this adds a 2-second Toast indicating that an ID has been copied successfully.

Screenshot 2024-12-06 at 10 24 22 AM

Delete
* this adds a function that can be greatly simplified once we refactor to just delete the specified row by query_id
* this adds a modal prompt alerting the user that this action cannot be undone and affects all users
* this adds a 2-second toast indicating that the specified query has been deleted
* if all queries are deleted, it reverts back to the emptyState page

Screenshot 2024-12-06 at 10 24 33 AM
Screenshot 2024-12-06 at 10 24 41 AM

Edit
* Still non-functional while we work on details of the edit-value-set workflow as well as the database refactor. Ideally, clicking the row or edit button will pull the `query_data` from the backend and render it on the edit-value-set page.

## Related Issue

Fixes #136, partially

## Additional Information
I leaned toward moving the Modal and Toast code to a `utils`. I had some issues with the [existing code](https://github.com/CDCgov/dibbs-query-connector/tree/main/query-connector/src/app/query/designSystem) for either the Modal or redirectToast working on this page. 

The existing Modal does not currently accept more button use cases; we could probably get away with just have `Delete`, but I took a stab at expanding to allow multiple buttons.

The Toast I was unable to get to render at all, unfortunately. The code for the toast is fairly lightweight; however, it could be worth debugging just to avoid pain down the line. At the very least, it probably makes sense to move the copy just to make that easier to update, as this text seems liable to change.

@mikang , a few questions about the above:
* Do we want another way besides a Toast to confirm success for Copy/Delete?
* Do we want another ways besides a Modal to prompt a user for a secondary confirmation that they wish to Delete?
* What changes should be made to the copy for each? I feel like I saw 

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation


mikang:
@robertandremitchell 
1) toast is good! it suffices
2) modal is great
3) I like the copy!

I'm not sure this is in the scope of this PR — could we vertically center these elements?

image

robertandremitchell:
> @robertandremitchell
> 
> 1. toast is good! it suffices
> 2. modal is great
> 3. I like the copy!
> 
> I'm not sure this is in the scope of this PR — could we vertically center these elements?
> 
> image

thanks! absolutely can take a look at this. I _think_ I fixed it with the css update by adding an `align-items: center`. let me know if this looks right, or if I've misunderstood!

query-connector/src/app/database-service.ts
```diff
@@ -829,6 +829,47 @@ export async function getCustomQueries(): Promise<CustomUserQuery[]> {
...
+    );
+  `;
+  const deleteQuerySql2 = `
+    DELETE FROM query_to_valueset WHERE query_id = $1;
+  `;
+  const deleteQuerySql3 = `
+    DELETE FROM query WHERE id = $1;
+  `;
+
+  const client = await dbClient.connect();
```
fzhao99, line 852: is there a reason we're making a new client connection and not just using the one that we instantiate earlier in this file?

fzhao99: think we might also be getting to the point that we start splitting up this file since it's getting pretty big

query-connector/src/app/queryBuilding/dataState/UserQueriesDisplay.tsx
```diff
@@ -15,25 +26,35 @@ interface UserQueriesDisplayProps {
...
-  };
+  const modalRef = useRef<ModalRef>(null);
+  const [selectedQuery, setSelectedQuery] = useState<{
+    queryName: string;
+    queryId: string;
+  } | null>(null);
 
   return (
     <div>
+      <ToastContainer position="bottom-left" />
```
fzhao99, line 43: We should probably put this into the `page.tsx` component should any of the subsequent pages in this route need toasts. Think the reason why we couldn't get the existing code housed in design system is because this higher-order wrapper wasn't there.

For a follow up PR: we should probably have a single page component across all our existing / future routes that pages can inherit from (that would include things like this ToastContainer, stylings for default page widths, etc.)

query-connector/src/app/queryBuilding/dataState/utils.tsx
```diff
@@ -0,0 +1,154 @@
...
+ * Renders a modal to confirm the deletion of a user query.
+ * @param modalRef - Reference to the modal component.
+ * @param selectedQuery - The currently selected query for deletion.
+ * @param handleDelete - Function to handle the deletion workflow.
+ * @param queries - The current list of user queries.
+ * @param setQueries - Function to update the state of queries.
+ * @param context - The data context used to update shared state.
+ * @returns The JSX element for the modal.
+ */
+export const renderModal = (
```
fzhao99, line 105: could we move this code into the `Modal.tsx` file (maybe renaming this method to be `renderDeletionModal` / tweaking the inputs to be a bit more generic) then export it to be consumed in the places where we need modals to be rendered? 

That way, the UI code that controls the look and styling of our modals are centralized and we can share this work further down the line in other places where we need modals

query-connector/src/app/database-service.ts
robertandremitchell: I think it was getting a little fussy when I was first working with the multiple deletions, but yeah, I can look into fixing this.

robertandremitchell: agreed on splitting out the file. I can create a tech debt ticket to address.

mikang:
Could we put the text in the same line as the icon?

image

query-connector/src/app/query/designSystem/redirectToast/RedirectToast.tsx
katyasoup, line 1: should we rename the file since it's not just a redirect anymore?

query-connector/src/app/database-service.ts
robertandremitchell: https://linear.app/skylight-cdc/issue/QUE-108/break-up-database-servicests-to-be-more-manageable

katyasoup (commented):
Looks great! Couple design things I spotted, but shouldn't block anything:

On my browser the toast text looks disproportionately large/different from your screenshots in the PR description; maybe just needs a different headingLevel passed in? 
Screenshot 2024-12-09 at 15 39 33  
 
 

I didn't see anything in Figma for the confirmation modal, but the serif font in the modal header doesn't match the other serif text on the page:
Screenshot 2024-12-09 at 15 45 34

query-connector/src/app/queryBuilding/dataState/utils.tsx
robertandremitchell: makes sense. I moved `Modal` and `DeleteModal` into a new folder. lmk your thoughts!

query-connector/src/app/queryBuilding/dataState/UserQueriesDisplay.tsx
robertandremitchell: moved the toast to `page.tsx` and created follow-up ticket: https://linear.app/skylight-cdc/issue/QUE-109/update-routing-for-formattingreused-components

robertandremitchell:
> Looks great! Couple design things I spotted, but shouldn't block anything:
> 
> On my browser the toast text looks disproportionately large/different from your screenshots in the PR description; maybe just needs a different headingLevel passed in? Screenshot 2024-12-09 at 15 39 33
> 
> I didn't see anything in Figma for the confirmation modal, but the serif font in the modal header doesn't match the other serif text on the page: Screenshot 2024-12-09 at 15 45 34


Thanks!

re: the different toast, this is expected based on @fzhao99 's work here https://github.com/CDCgov/dibbs-query-connector/pull/204#event-15588700420 to leverage/match the existing toast for the demo customize query work?

@mikang , are you still okay with the presentation of the toast?

for the modal header, I've updated the .scss to the appropriate font for modal header. We may need to revisit if the Modals end up differing on header font/formatting, but as the inaugural Modal...uh, first come, first serve? 😅

query-connector/src/app/query/designSystem/redirectToast/RedirectToast.tsx
robertandremitchell: maybe. @fzhao99 , are you good if we just rename it to `Toast`? it seems pretty flexible to cover our current use cases.

katyasoup (approved):

query-connector/src/app/query/designSystem/redirectToast/RedirectToast.tsx
fzhao99: sounds good to me!

fzhao99 (approved):
lgtm!

Issue #190, opened 2024-12-05
Populate pg fhir server table with one value


bamader:
## Description:

Use flyway insert statements to pre-populate the new table with the information for the Public HAPI server. The thinking here is that this server is always good to have around for demo purposes and has the least overhead config to set up to get the vertical slice up and running.

Additionally, write a function in database-service.ts called \`getFhirServerConfig\` that retrieves the configuration information for a FHIR server by the server’s name.

Finally, change the \`fhirServers\` constant in fhir-servers to pull the information for the Public HAPI server from the DB.

## Acceptance Criteria:

1. The flyway migration that creates the table has pre-populated data for one row.
2. A getFhirServerConfig function is written.
3. The fhirServers constant is modified to load from the DB rather than hard code

PR #203, opened 2024-12-09
Add `fhir_servers` table + `getFhirServerConfigs` function

### LINKED ISSUE #189: Create a PG Table to store FHIR server connection config details
## Description:

This table can be created in a flyway migration that follows query and valueset table creation. The table needs to have a number of columns for storing credentials:

1. id: generic UUID column for the DB
2. server_name: what to call the server (e.g. Public HAPI Server)
3. hostname: the endpoint URL associated with the FHIR server
4. headers: an optional JSON column containing information such as content-type, accept, cache-control, and the like

## Acceptance Criteria:

A table is created matching the above column specification

m-goggins:
# PULL REQUEST

## Summary

This PR adds the initial step to pull FHIR server configurations from the DB and run them using configuration from the DB instead of relying on hardcoded values. 

The new Flyway migration builds out the `fhir_servers` table where connection details will be stored. It also adds a single row to the `fhir-servers` table that contains the details for the Public HAPI connection.

I also added a function called `getFhirServerConfigs` to retrieve and use the FHIR server configuration(s) when executing a query as well as `getFhirServerNames` to retrieve just the names for rendering on the client side without exposing config details. 

I had to do a little finagling with how we handle the `init` to get both the hardcoded server names and the server names in the flyway migration to appear and run because the configuration for the hardcoded values was set up differently from how we designed the table.

## Related Issue

Fixes #189 + [#190](https://linear.app/skylight-cdc/issue/QUE-97/populate-pg-fhir-server-table-with-one-value)

## Additional Information
I'm going to make a follow up ticker/PR to resolve how we're handling the `FHIR_SERVERS` const on the API side but figured that this PR is the bulk of the backend work. 

## Checklist

- [x] Link to relevant issues
- [x] Update documentation


robertandremitchell (approved):
broadly looks good to me. I wasn't able to test the demo workflow entirely because the public HAPI server is down, but advanced workflows still worked.

query-connector/src/app/database-service.ts
```diff
@@ -885,3 +886,42 @@ export async function checkDBForData() {
...
+
+//Cache for FHIR server configurations
+let cachedFhirServerConfigs: Promise<FhirServerConfig[]> | null = null;
+
+/**
+ * Fetches all FHIR server configurations from the database and caches the result.
+ * @returns An array of FHIR server configurations.
+ */
+export async function getFhirServerConfigs() {
+  if (cachedFhirServerConfigs) {
```
fzhao99, line 898: suggestion to be taken with some salt: would we ever want an escape hatch (via an optional param?) to flush the cache?

fzhao99 (approved):
left one suggestion but lgtm!

query-connector/flyway/sql/V02_01__add_fhir_servers.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS fhir_servers (
+    id UUID DEFAULT uuid_generate_v4 (),
+    name TEXT,
+    hostname TEXT,
+    headers JSON DEFAULT NULL
```
nickclyde, line 5: Will we need to add columns for authentication creds for non-public servers, or will we just store a token in the headers column?

query-connector/src/app/database-service.ts
nickclyde: Since this is a server action, I think we won't be able to flush the cache just by refreshing the page, so I think Bob is right that we'll need some manual way to flush the cache

query-connector/flyway/sql/V02_01__add_fhir_servers.sql
m-goggins: I believe we'll just store them in the headers column unless you think there's a good reason to keep it separate. We have some OAUTHSCOPES in the headers for the CernerHelios connection but we haven't had any username, etc. authentication yet.

```
function configureEHX(xdestination: string): DevFhirServerConfig {
  const headers = {
    Accept: "application/json, application/*+json, */*",
    "Accept-Encoding": "gzip, deflate, br",
    "Content-Type": "application/fhir+json; charset=UTF-8",
    "X-DESTINATION": xdestination,
    "X-POU": "PUBHLTH",
    "X-Request-Id": uuidv4(),
    prefer: "return=representation",
    "Cache-Control": "no-cache",
  };
  if (xdestination === "CernerHelios") {
    (headers as Record<string, string>)["OAUTHSCOPES"] =
      "system/Condition.read system/Encounter.read system/Immunization.read system/MedicationRequest.read system/Observation.read system/Patient.read system/Procedure.read system/MedicationAdministration.read system/DiagnosticReport.read system/RelatedPerson.read";
  }
  return {
    id: xdestination,
    name: xdestination,
    hostname: "https://concept01.ehealthexchange.org:52780/fhirproxy/r4/",
    headers,
    trustSelfSigned: true,
  };
}
```

query-connector/src/app/database-service.ts
m-goggins: Added a `forceRefresh` param that we can set true to manually reset. I'm thinking it might be nice to refresh when we [add/edit/delete fhir servers](https://linear.app/skylight-cdc/issue/QUE-100/add-edit-and-delete-buttons-to-front-end-fhir-server-table).

PR #214, opened 2024-12-12
fix translation methods between query pages

fzhao99:
# PULL REQUEST

## Summary
The code that just went in for the new DB tables overloaded a mapping method between @katyasoup and I's PR's that should have been two separate functions. This PR separates them and adds some unit tests to prevent this in the future

## Additional Information
Another example of us not having testing infra biting us! Will bring this up at retro

Something else that might have caught this earlier would be more explicitly typing the response from the DB beyond the provided `QueryRow`. **Would love opinions on how we might type our queries more explicitly to tell application code what the shape of data coming back from the DB looks like**

## Checklist
- [ ] Update documentation

### DIFF
--- a/query-connector/src/app/database-service.ts
+++ b/query-connector/src/app/database-service.ts
@@ -688,8 +688,7 @@ export async function getCustomQueries(): Promise<CustomUserQuery[]> {
       q.query_data,
       q.conditions_list
     FROM
-      query q
-    WHERE     q.query_name IN ('Gonorrhea case investigation', 'Newborn screening follow-up', 'Syphilis case investigation', 'Cancer case investigation', 'Chlamydia case investigation');
+      query q;
   `;
   // TODO: this will eventually need to take into account user permissions and specific authors
   // We'll probably also need to refactor this to not show up in any user-facing containers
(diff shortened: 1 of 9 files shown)


query-connector/src/app/database-service.ts
```diff
@@ -688,8 +688,7 @@ export async function getCustomQueries(): Promise<CustomUserQuery[]> {
       q.query_data,
```
fzhao99, line 688: we can just return all the entries in the query table now that we only seed the table with the five demo queries

query-connector/src/app/utils.tsx
```diff
@@ -76,39 +76,94 @@ type QueryTableQueryDataColumn = {
...
+    author: storedConcept["author"],
+    system: storedConcept["code_system"],
+    ersdConceptType: storedConcept["type"] ? storedConcept["type"] : undefined,
+    dibbsConceptType: storedConcept["dibbs_concept_type"],
+    includeValueSet: conceptGroup.find((c) => c["include"]) ? true : false,
+    concepts: conceptGroup.map((c) => {
+      return {
+        code: c["code"],
+        display: c["display"],
+        include: c["include"] ?? true,
```
fzhao99, line 160: I think the contexts where `include` isn't defined off the concept group, (I think?) the default when we're making a trip back from the DB should be true.

robertandremitchell (approved):

Issue #169, opened 2024-11-20
Update navigation between queryBuilding pages


robertandremitchell:
## Description:

Now that we have the backend / db schema work finished and the  independent pages for `My queries` and the beginnings of the building screens, we'll need to link them together. This includes the following: 

* Once work is complete to add `query_data` and `conditions_list` to the `query` table in the database, we will need to refactor this section of code: [https://github.com/CDCgov/dibbs-query-connector/blob/main/query-connector/src/app/database-service.ts#L622](https://github.com/CDCgov/dibbs-query-connector/blob/main/query-connector/src/app/database-service.ts#L622)
  * We should be able to avoid as many joins and just pull data from `query`. This may require updating the interface to not pull from valuesets or to reflect the JSON we create for `query_data`.
  * This should allow us to also in the case of superadmins get rid of any WHEREs. which only exist now to speed up processing. The limiting factor should be done based on user permissions, which we may need to add here or could potentially be accomplished in an alternative method.
* Refactoring the navigation to link between the my queries page and the building page
* Updating the fetch to pull from the new `query_data` column
* Cleaning up any unused code

## Acceptance Criteria:

* Fetches valuesets / concepts for conditions pulls from the `query_data` column
* My Queries table pulls from the `query` table rather than hardcoded
* Constants that tie the demo query options to labels / values / etc aren't so duplicative

PR #217, opened 2024-12-16
update navigation between query building pages

fzhao99:
# PULL REQUEST

## Summary
Links the various query building work (backend schema / fetching functionality, the My Queries page, and the valueset selection page) together

- Changes the query building pages to use the frontend-driven navigational style that exists in query customization
    - next router (which is what we were using) doesn't have a default way of persisting state between pages. I tried hacking together something together using context, but it didn't work great / caused issues, so I went ahead and did it like we're doing in the other flow
    - Something worth investigating into the future is a more unified approach into our router setup. The way we're currently handling routing doesn't play nice with the browser back button, which has accessibility issues. Am going to do a bit of reading on the differences between navigation options in NextJS and write up a ticket.
 - Populate the valueset selection page with the query name / selected valuesets of saved queries 
 - Also tweaked the population of the `include` prop for valuesets to match closer to what we've discussed. This fixes a regression for the valueset accordions being checked on by default that got introduced
 - Started splitting out new additions to the backend into their separate files, first by moving the DB client creation to another file

## Related Issue
Part one of #169 

https://github.com/user-attachments/assets/abc8de3f-912e-48ae-b615-8092ea5cc34b

## Additional Information
- This should give an easier way for us to reference / update which valuesets/conditions we're associating with queries we're building, since the ID / name are now passed between all these pages
- Followup cleanup PR incoming to dedupe some constants

## Checklist
- [x] Link to relevant issues

### DIFF
--- a/query-connector/src/app/constants.ts
+++ b/query-connector/src/app/constants.ts
@@ -404,3 +404,11 @@ export const INTENTIONAL_EMPTY_STRING_FOR_CONCEPT_VERSION = "";
 // we clean up the DB migration. Leaving these in until we can clean these up
 // in the migration schema
 export const INTENTIONAL_EMPTY_STRING_FOR_GEM_CODE = "";
+
+// Define the type for the FHIR server configurations
+export type FhirServerConfig = {
+  id: string;
+  name: string;
+  hostname: string;
+  headers: Record<string, string>;
+};
--- a/query-connector/src/app/constants.ts
+++ b/query-connector/src/app/constants.ts
@@ -24,12 +24,12 @@ export type USE_CASES = (typeof UseCases)[number];
 export const UseCaseToQueryName: {
   [key in USE_CASES]: string;
 } = {
-  "social-determinants": "Social Determinants of Health",
-  "newborn-screening": "Newborn Screening",
-  syphilis: "Congenital syphilis (disorder)",
-  gonorrhea: "Gonorrhea (disorder)",
-  chlamydia: "Chlamydia trachomatis infection (disorder)",
-  cancer: "Cancer (Leukemia)",
+  "social-determinants": "Gather social determinants of health",
+  "newborn-screening": "Newborn screening follow-up",
+  syphilis: "Syphilis case investigation",
+  gonorrhea: "Gonorrhea case investigation",
+  chlamydia: "Chlamydia case investigation",
+  cancer: "Cancer case investigation",
 };
 
 /**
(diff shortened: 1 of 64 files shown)


query-connector/src/app/queryBuilding/buildFromTemplates/BuildFromTemplates.tsx
```diff
@@ -88,13 +117,6 @@ export default function QueryTemplateSelection() {
       const formattedResults =
         results && groupConditionConceptsIntoValueSets(results);
 
-      // when fetching directly from conditions table (as opposed to a saved query),
```
fzhao99, line 91: I went ahead and changed the default return from the formatting function to have the `include` values to be true, since in any case that I can think of, fetching a valueset / concept means that we want to use it. Lmk if this is an incorrect assumption though.

robertandremitchell: I think that's right for a default condition/valueset/concept tree that we'd include everything. I also like the code to deselect a valuset if someone deselects all concepts.

query-connector/src/app/constants.ts
```diff
@@ -300,7 +300,7 @@ export const stateOptions = [
 export type Mode = "search" | "results" | "select-query" | "patient-results";
 
 /* Mode that query building pages can be in; determines what is displayed to the user */
-export type BuildStep = "condition" | "valueset" | "concept";
+export type BuildStep = "selection" | "condition" | "valueset" | "concept";
```
robertandremitchell, line 303: makes sense. I don't think we actually need a `concept` BuildStep, since concepts are selected as part of the Drawer component on the valueset page.

robertandremitchell (approved):
looks good! definitely agree that keeping the same structure as we do for demo makes the most sense for maintaining state.

query-connector/src/app/queryBuilding/components/SelectionTable.tsx
```diff
@@ -112,11 +112,11 @@ export const SelectionTable: React.FC<SelectionTableProps> = ({
...
           groupedValueSetsForCondition[valueSetType],
         );
-        const totalCount = tallyConcpetsForValueSetGroup(
+        const totalCount = tallyConceptsForValueSetGroup(
           valueSetsForType,
           false,
         );
-        const selectedCount = tallyConcpetsForValueSetGroup(
+        const selectedCount = tallyConceptsForValueSetGroup(
           valueSetsForType,
```
fzhao99, line 120: fixed this typo

robertandremitchell (approved):
Broadly looks good. found a few more lines in need of cleanup with the retooling, but should be good to merge after.

* https://github.com/CDCgov/dibbs-query-connector/blob/d52e3b1b6f754927d083899a0b79eafff42014f4/query-connector/src/app/queryBuilding/querySelection/EmptyQueriesDisplay.tsx#L28
* https://github.com/CDCgov/dibbs-query-connector/blob/d52e3b1b6f754927d083899a0b79eafff42014f4/query-connector/src/app/query/components/header/header.tsx#L145

PR #224, opened 2024-12-17
clean up duplicative query type constants

fzhao99:
Follow up for this comment: https://github.com/CDCgov/dibbs-query-connector/pull/201#discussion_r1880649603

### DIFF
--- a/query-connector/src/app/constants.ts
+++ b/query-connector/src/app/constants.ts
@@ -8,69 +8,35 @@ import {
   MedicationAdministration,
   MedicationRequest,
 } from "fhir/r4";
-/**
- * The use cases that can be used in the app
- */
-export const UseCases = [
-  "social-determinants",
-  "newborn-screening",
-  "syphilis",
-  "gonorrhea",
-  "chlamydia",
-  "cancer",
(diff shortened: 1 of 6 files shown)


query-connector/src/app/constants.ts
```diff
@@ -8,69 +8,35 @@ import {
...
-  { value: "syphilis", label: "Syphilis case investigation" },
-];
 
-type DemoQueryOptionValue = (typeof demoQueryLabels)[number];
-export const demoQueryValToLabelMap = demoQueryOptions.reduce(
-  (acc, curVal) => {
-    acc[curVal.value as DemoQueryOptionValue] = curVal.label;
-    return acc;
+export const USE_CASE_DETAILS = {
+  "social-determinants": {
```
robertandremitchell, line 13: I don't think we want to actually include social determinants anymore; this was a more custom query that we haven't fully supported as of yet (so if they try to customize it, they get 0 things to customize). Or we could comment it out for now to hedge.

robertandremitchell (approved):
looks good! so much more intuitive. just a small note about social determinants

PR #229, opened 2024-12-19
Query Updating Concept Selection 

robertandremitchell:
# PULL REQUEST

## Summary

This finishes the frontend implementation of the Drawer component for the View Codes concept workflow . It adds:
* A dynamic pulling of the concept code and name to display when view codes is clicked
* A super-checkbox for the entire valueset
* Checkboxes for each of the concepts
* A Save Changes button that is disabled until a change from the status quo is detected
* A warning Modal if there are changes to ask if the changes should be saved or dismissed

Screenshot 2024-12-19 at 8 54 29 AM
Screenshot 2024-12-19 at 8 54 40 AM
Screenshot 2024-12-19 at 8 54 46 AM
Screenshot 2024-12-19 at 8 54 55 AM
Screenshot 2024-12-19 at 8 55 10 AM

## Related Issue

Fixes #140

## Additional Information

I'm having some trouble writing the code to get the changes to "stick". I.e., if there are six concepts, we remove one, we should see `5/6` for the valueset. I have not yet been able to get this work / the changes to truly save on a concept-by-concept level. Gonna keep hacking at it this morning.

Happy to pair on it, or tackle it as part of a larger ticket on getting every part of the page to save, as that is currently nonfunctional cc @fzhao99.

@mikang , I am curious if we want a the Minus checkbox in the concept workflow? I.e., if one of six concepts is deselected, the super-checkbox would display the minus-sign checkbox. Right now, in the designs, the checkbox is still selected, which I guess kinda makes sense given that we are still included 1 whole valueset, just with concepts missing. But I wasn't sure. As always, if there are any other frontend display changes/copy to modify/add lmk!

Anything else the review team should know?

## Checklist

- [ ] Descriptive Pull Request title
- [ ] Link to relevant issues
- [ ] Provide necessary context for design reviewers
- [ ] Update documentation

Issue #140, opened 2024-11-11
Implement query updating on concept selection


fzhao99:
## Description:

Based on the determination from [QUE-70](https://linear.app/skylight-cdc/issue/QUE-70/spike-when-how-to-do-the-valueset-update-to-the-backend), implement the "write" part of the slide over

## Acceptance Criteria:

* A user can select and save individual concepts from the query building concept selection flow

PR #231, opened 2024-12-20
refactor valueset translation types and functions

fzhao99:
# PULL REQUEST

## Summary
During the work for https://github.com/CDCgov/dibbs-query-connector/pull/229, ran into some issues with the way we're organizing our ValueSet translation code / types. Decided it'd be worth our time to refactor these to be more clear and picked up some other random cleanup steps as well.
 
- The bulk of the interesting work is happening in `utils/valueSetTranslation` - moved a bunch of translation function code into this file that's shared between query customization and building
     - Will go back and write some tests for these utilities since they're begging to be the source of subtle bugs  
- Renamed our internal ValueSet --> `DibbsValueSet` to prevent confusion over the one provided by FHIR
- Renamed a bunch of our mapping types using the `[GrandParentKey]To[ParentKey]To[ParentValue]Map` standard to hopefully make a bit more explicit what the multi-layer maps are doing
    - These types are admittedly pretty gross, but think they represent uncompressible complexity in cases where we have grandparent - parent - child maps that are holding category - condition - valueset - concept hierarchy. If anyone has thoughts on how to simplify though, am all ears! 
- In a few places, we were mixing up `DibbsValueSet` with what we used to call `ValueSetGroups`. 
     - The latter is the type we introduced in query customization is the collation of all VS's that share author / system / name
     - Renamed this to `VsGrouping` to hopefully make this a bit more clear
- Moved the db creation script into its own directory within `backend`, and moved a test for the translation function / the function itself into separate files
- Updated imports / references
- Cleaned up a few other random things

## Related Issue
Setup for #140 since the naming conventions were causing confusion over that branch of the codebase / some functions were being reused from query customization

## Additional Information
Will follow up with 
- A PR that tests the code we moved into `utils/valueSetTranslation`
- A rebased version of https://github.com/CDCgov/dibbs-query-connector/pull/229 that'll use these new types

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation


query-connector/src/app/query/components/customizeQuery/customizeQueryUtils.ts
fzhao99, line 1: split out this file into other files

query-connector/src/app/database-service.ts
```diff
@@ -197,48 +190,12 @@ export async function getVSACValueSet(
   }
 }
 
-/**
- * Translates a VSAC FHIR bundle to our internal ValueSet struct
- * @param fhirValueset - The FHIR ValueSet response from VSAC
- * @param ersdConceptType - The associated clinical concept type from ERSD
- * @returns An object of type InternalValueSet
- */
-export async function translateVSACToInternalValueSet(
```
fzhao99, line 206: moved this out to its own, non-server component file so that we didn't have to mark it as async to export it

query-connector/src/app/query/components/CustomizeQuery.tsx
```diff
@@ -253,3 +251,20 @@ const CustomizeQuery: React.FC<CustomizeQueryProps> = ({
 export default CustomizeQuery;
 export const QUERY_CUSTOMIZATION_CONFIRMATION_BODY =
   "Query customization successful!";
+
```
fzhao99, line 254: moving this over from a deleted file

query-connector/src/app/utils/valueSetTranslation.ts
```diff
@@ -0,0 +1,162 @@
...
+ * Helper function that takes an array of value set items and groups them using
+ * a combination of the value set name, author, and system to create a unique
+ * grouping of items. These groups are displayed as individual accordions on
+ * the customize query page
+ * @param valueSetsToGroup - an array of value sets to group
+ * @returns - a dictionary of value sets, where the index are the unique combinations
+ * of valueSetName:author:system and the values are all the value set items that
+ * share those identifiers in common, structed as a GroupedValueSet
+ */
+export function groupValueSetsByNameAuthorSystem(
```
fzhao99, line 32: this method was moved over from another file

query-connector/src/app/utils/valueSetTranslation.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/**
+ * Helper function to map an array of value sets into their DibbsConceptType
+ * buckets
+ * @param valueSets - A list of value sets
+ * @returns Dict of list of rows containing only the predicate service type
+ * mapped to one of the DibbsConceptTypes.
+ */
+export const groupValueSetsByConceptType = (
```
fzhao99, line 147: as was this one

query-connector/src/app/queryBuilding/components/SelectionTable.tsx
```diff
@@ -56,13 +60,13 @@ export const SelectionTable: React.FC<SelectionTableProps> = ({
...
-    valueSetType: DibbsValueSetType,
-    groupedValueSet: GroupedValueSet,
+    activeConceptType: DibbsConceptType,
+    groupedValueSet: VsGrouping,
     isBatchUpdate: boolean = false,
     batchValue?: boolean,
   ) => {
     const key = `${groupedValueSet.valueSetName}:${groupedValueSet.author}:${groupedValueSet.system}`;
-    const updatedVS = groupedValueSetsForCondition[valueSetType][key];
+    const updatedVS = groupedValueSetsForCondition[activeConceptType][key];
```
robertandremitchell, line 69: a bit confused by the naming of this to `activeConceptType`. Aren't these valueset-related, not concept-related?

robertandremitchell (commented):
I think this all broadly makes sense, but I'm a bit confused about the new `conceptType` renaming. I don't know what name makes the most sense here, but if I understand it correctly, this is supposed to be lab/conditions/medications for a valueset.

Since our hierarchy is category -> condition -> valueset -> concept, it could be a bit confusing, even though at least in the eRSD we do determine the valueset type via concept-level data, so I get the renaming, but I do find `valueSetType` to be a tad more intuitive in indicating what level of data we're currently working with.

query-connector/src/app/queryBuilding/components/SelectionTable.tsx
fzhao99: good point! went ahead and changed this back

robertandremitchell (approved):

PR #242, opened 2025-01-03
customize query - graceful fallback for empty response

fzhao99:
# PULL REQUEST

## Summary
Adding a graceful fallback so that there aren't client-side errors for the db issues we've been running into. True fix still TBD.

https://github.com/user-attachments/assets/833e26f4-0992-4da6-8439-3b482c1d240c

## Checklist


robertandremitchell (approved):
makes sense, lgtm

PR #248, opened 2025-01-07
FHIR Server Configuration

nickclyde:
# PULL REQUEST

## Summary

Adds a page at `/fhir-servers` to add and edit FHIR servers. Based on [these designs](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=1155-4577&m=dev).

Currently resolves the following tickets:
https://linear.app/skylight-cdc/issue/QUE-98/create-basic-front-end-for-fhir-server-config
https://linear.app/skylight-cdc/issue/QUE-100/add-edit-and-delete-buttons-to-front-end-fhir-server-table
https://linear.app/skylight-cdc/issue/QUE-101/create-fhir-server-front-end-popup-for-create-new
https://linear.app/skylight-cdc/issue/QUE-102/write-a-creditedit-fhir-server-config-function
https://linear.app/skylight-cdc/issue/QUE-103/write-a-deletefhirserver-function-to-remove-a-db-config
https://linear.app/skylight-cdc/issue/QUE-104/finish-migrating-hard-coded-fhir-servers
https://linear.app/skylight-cdc/issue/QUE-113/update-fhirservers-and-fhir-servers

Screenshots:

Screenshot 2025-01-09 at 12 29 05 PM
Screenshot 2025-01-09 at 12 28 32 PM
Screenshot 2025-01-09 at 12 28 42 PM
Screenshot 2025-01-09 at 12 29 20 PM
Screenshot 2025-01-09 at 12 29 36 PM


query-connector/src/app/fhir-servers/fhirServers.module.scss
```diff
@@ -0,0 +1,62 @@
...
+  font-size: 0.875rem;
+}
+
+.tableRowHover {
+  transition: background-color 0.2s ease;
+  border-collapse: collapse;
+
+  &:hover,
+  &:hover > td {
+    background-color: rgba(217, 232, 246, 0.50);
```
katyasoup, line 39: you should be able to apply the background to just the `&:hover` parent, and then the `td` children will inherit it (see [here](https://stackoverflow.com/questions/27332016/how-do-i-layer-css-opacity) for more of an explanation)

that seems to be working when I adjust it on my local, but I know I ran into a similar issue before where the child elements weren't picking it up, and I had to do some math-y magic (like, parent has 0.5, child has 0.125) - lmk if the above doesn't do the trick!

nickclyde: Amazing, this worked!! Thanks @katyasoup!!

katyasoup (commented):
Looking good! Couple design things I noticed (ignore if these are out of scope!)

Screenshot 2025-01-08 at 14 18 48

- the Delete button looks like it should be left-justified per the Figma mockup

Screenshot 2025-01-08 at 14 18 55

- when I click Test Connection and get a success result, the buttons all grow in height; looks like Figma only shows the success response without the Delete button so I'd double check with @mikang on that
- "Save Changes" always results in a "Failed to update the server configuration." error (but maybe that is out of scope?)

mikang:
Thanks @katyasoup !

- The cancel button should be just a link, not a white button. Figma below:
image

- Actually, that's a good catch. The Delete button should be there if the user is editing and then testing the connection. Can we keep the Delete button and change the text to "[checkmark] Success!" ? That shortens the text, so the buttons don't grow in height.

m-goggins:
Would you be able to add the tickets that this PR covers? It looks like a lot of the ones written up in December but it'd be helpful to see what is left before the Connectathon.

nickclyde:
@katyasoup @mikang Just pushed the design fixes for the stuff you pointed out!

@m-goggins added the list of tickets this addresses so far. I plan on implementing delete functionality next and moving the hardcoded servers into the table after that as discussed at standup!

robertandremitchell (commented):
looks good in `npm run dev`! I was able to successfully add and remove a fhir server without issue.

running into this error on `npm run build`

```

./src/app/fhir-servers.ts:14:3
Type error: Type '{ id: string; name: string; hostname: string; headers: {}; }' is not assignable to type 'DevFhirServerConfig'.
  Type '{ id: string; name: string; hostname: string; headers: {}; }' is missing the following properties from type 'FhirServerConfig': last_connection_attempt, last_connection_successful

  12 |   process.env.E2E_LOCAL_FHIR_SERVER ?? "http://hapi-fhir-server:8080/fhir";
  13 | export const fhirServers: Record<string, DevFhirServerConfig> = {
> 14 |   "HELIOS Meld: Direct": {
     |   ^
  15 |     id: "HELIOS Meld: Direct",
  16 |     name: "HELIOS Meld: Direct",
  17 |     hostname: "https://gw.interop.community/HeliosConnectathonSa/open",
```

fzhao99 (approved):
this lgtm! think the build error is a pretty simple fix, so approving.

Wondering if we should have a followup that adds our hardcoded default servers to the dump file? Might be good to have them available out of the box like we do the default queries.

query-connector/src/app/fhir-servers/page.tsx
```diff
@@ -0,0 +1,392 @@
...
+    success: boolean;
+    error?: string;
+  }
+
+  /**
+   * Tests a connection to a FHIR server
+   * @param url - The URL of the FHIR server to test
+   * @returns A promise that resolves to an object containing the test result
+   */
+  const testFhirConnection = async (
```
m-goggins, line 78: Definitely could be a separate ticket/post-connectathon but I think we should try to test the Connection upon insertion, at least for the default servers, so that when a user first looks at the table the status is not null.

m-goggins (approved):
Looks good to me as long as we update the table to pull the list of available FHIR servers from the table and the hard coded list like we do for the drop down or update the table to include all of the hard coded servers (even is just for the connectathon). Left a few other comments for consideration but nothing that should be blocking.

query-connector/src/app/fhir-servers/page.tsx
```diff
@@ -0,0 +1,392 @@
...
+        onClick: handleDeleteServer,
+      });
+    }
+
+    // Update test connection button if successful
+    if (connectionStatus === "success") {
+      buttons[1].text = (
+        <>
+          <Icon.Check
+            size={3}
```
katyasoup, line 256: I'm still seeing the button size jump after the connection check, which I think is caused by the checkmark icon? Bumping this down to size 2 keeps it in line with the line height of the text

query-connector/src/app/fhir-servers/page.tsx
```diff
@@ -0,0 +1,392 @@
...
+            className="usa-icon"
+            aria-label="Connected"
+            color="green"
+          />
+          Success
+        </>
+      );
+      buttons[1].className =
+        "usa-button usa-button--outline shadow-none text-green padding-left-0 padding-right-2";
```
katyasoup, line 265: can we add the same left padding here so the checkmark + text are centered evenly?

query-connector/src/app/fhir-servers/page.tsx
katyasoup: (or decrease the top/bottom padding to accommodate, since I see the linter yelling about 2 not being an acceptable value here)

nickclyde:
Just fixed the build by moving all the hardcoded servers into the database via migration. Also cleaned up the `FhirServers` constant references. Also had to make the import of `Modal` a dynamic import to make the build work. Still have a few comments to address but I think this now covers almost all of the FHIR server configuration tickets in Linear!

robertandremitchell (approved):
Looks good! a few formatting touch-ups, but I think they would be fine to tackle in a follow-up ticket if we want to get the bulk of this merged.

Screenshot 2025-01-13 at 8 28 30 AM

https://github.com/user-attachments/assets/c06e6fdb-746c-41e1-9ec6-ca5bf813ca3a

query-connector/src/app/fhir-servers/page.tsx
```diff
@@ -0,0 +1,392 @@
...
+        let errorMessage: string;
+        switch (response.status) {
+          case 401:
+            errorMessage =
+              "Connection failed: Authentication required. Please check your credentials.";
+            break;
+          case 403:
+            errorMessage =
+              "Connection failed: Access forbidden. You do not have permission to access this FHIR server.";
+            break;
```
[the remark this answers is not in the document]
nickclyde: Hmm, I'm not sure. We can always adjust later, but I think since everything else on the page calls it FHIR server let's just leave it for now

query-connector/src/app/fhir-servers/page.tsx
nickclyde: It actually already does this! See line 178

nickclyde:
Good catch @robertandremitchell! Fixed those issues!

PR #250, opened 2025-01-07
Modify application url to root

shanice-skylight:
Modify application url to root
# PULL REQUEST

## Summary

- Fix terraform to workflow by adding variables file in command and -replace flag so that each deployment the image is replace with latest image
- Deleted middleware.js file
- Deleted next.config.js

## Related Issue

Fixes #

## Additional Information

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation


robertandremitchell (commented):
broadly makes sense, but I think there's a few files like `playwright-setup.ts` that also need to be updated with the right URL. 

I _think_ that should fix the e2e tests that are timing out, though it's not working for locally, as my docker is being very stubborn in building images...

nickclyde:
Let's make sure we don't merge this until we have already pushed whatever code we want to freeze on dibbs.cloud, because this will break that environment if it makes it in!

nickclyde (approved):

PR #256, opened 2025-01-09
query building write functionality pt 1

fzhao99:
# PULL REQUEST

## Summary
Part one of the query building effort to slim down the final PR. Changes include

- We decided to store the JSON in the `query_data` and `conditions_list` columns of the query table by condition ID rather than condition name like we had it previously. Updated the fixtures / unit tests / seed file and one place in the frontend UI accordingly
- Introduce the code to do the query upserts, to be consumed in the upcoming PR
- Added some UI gating / updating: should be a no-op for now

## Related Issue

Part one of #255 

## Checklist

- [x] Link to relevant issues
- [ ] Update documentation

### DIFF
--- a/query-connector/src/app/backend/query-building.ts
+++ b/query-connector/src/app/backend/query-building.ts
@@ -1,7 +1,10 @@
 "use server";
 
 import { getDbClient } from "./dbClient";
-import { QueryDetailsResult } from "../queryBuilding/utils";
+import { NestedQuery, QueryDetailsResult } from "../queryBuilding/utils";
+import { DibbsValueSet } from "../constants";
+import { DEFAULT_TIME_WINDOW } from "../utils";
+import { randomUUID } from "crypto";
 const dbClient = getDbClient();
 
 /**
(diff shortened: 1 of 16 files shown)


query-connector/src/app/backend/query-building.ts
```diff
@@ -28,3 +31,87 @@ export async function getSavedQueryDetails(queryId: string) {
...
+      return result.rows as unknown as QueryDetailsResult[];
+    }
+    console.error("Query save failed:", dataToWrite);
+    return [];
+  } catch (error) {
+    console.error("Error saving new query", error);
+  }
+}
+
+function formatQueryDataForDatabase(frontendInput: NestedQuery) {
```
fzhao99, line 88: Decided to store the data in the `query_data` column as `conditionId > valueSetId > ValueSet`. Think this gives us the most flexibility for further manipulation in the backend, but can change this as needed

robertandremitchell (commented):
looks good. I'm having some issues running in dev mode, but I'm not sure if end to end is supposed to work. It looks like the edit querying is a little funky, but maybe it is addressed in part 2:

https://github.com/user-attachments/assets/f488bd4c-7e12-4576-bf6b-00f0f0be8c9e

fzhao99:
> looks good. I'm having some issues running in dev mode, but I'm not sure if end to end is supposed to work. It looks like the edit querying is a little funky, but maybe it is addressed in part 2:
> 
> 
> 
> https://github.com/user-attachments/assets/f488bd4c-7e12-4576-bf6b-00f0f0be8c9e
> 
> 
> 
> 

Yeah this part is just setup! Editing coming in part two

robertandremitchell (approved):

PR #269, opened 2025-01-13
Make keycloak work in docker compose

### LINKED ISSUE #252: Add feature flag around sign in button
## Description:

Currently, the sign in button on dibbs.cloud will cause an error since Keycloak is not deployed in that environment. We should wrap that button's handler function with an environment variable (eg AUTH_DISABLED) which, when enabled, just makes the button redirect to `/query`

## Acceptance Criteria:

When AUTH_DISABLED is true, the sign in button should just redirect to `/query`

Otherwise, the sign in button should attempt authentication with Keycloak as it does now.

bamader:
# Update Keycloak Authorization

## Summary

This PR covers two points:
(1) Adding a small wrapper in the header component that allows us to use an environment variable to disable authentication; this is a convenience for local development as well as enables us to turn off authentication in dev/prod when it's not doing anything.
(2) Reformulates the `auth.ts` file to more robustly and comprehensively link NextAuth with Keycloak. Previously, keycloak could not be spun up using `docker compose`, due to networking and connectively issues between the server of the app and the server for keycloak. This new configuration and suite of environment variables allows us as devs to work with keycloak and the app in both dev mode (`npm run dev`) and using docker compose, with updated port mappings and flags to boot.

## Related Issue

Fixes #252 

## Additional Information

We can't access `process.env` environment variables from a purely client side component (which Header is) because from its perspective, there is no process.env, so we just have to modify the component signature to accept whether auth is disabled rather than calculate for itself.


query-connector/docker-compose-dev.yaml
```diff
@@ -49,6 +49,8 @@ services:
       - hapi-fhir-server
 
   # Keycloak for authentication
+  # Note: you must have a local .env file with a parameter AUTH_DISABLED.
```
m-goggins, line 52: Should this be added to the .env.sample?

m-goggins (commented):
Everything spun up for me, thought when accessed the query connector page, I got this error in the logs. I'm not 100% sure what it means. Might have to do with some refactoring we did prior to the connectathon?

```
query-connector-1  | [auth][error] UntrustedHost: Host must be trusted. URL was: http://0.0.0.0:3000/api/auth/session. Read more at https://errors.authjs.dev#untrustedhost 
```

query-connector/docker-compose-dev.yaml
bamader: Good catch, will add that!

bamader:
@m-goggins Hmm I'm not sure where that's coming from, especially since we have the TRUST_HOST flag set in `auth.ts` and `REJECT_UNAUTHORIZED=0` in the docker compose file. I'll see if I can repro.

nickclyde (approved):
Thanks Brandon!!

bamader:
@m-goggins I can't manage to repro that log error at all actually :/ . If it's not blocking you and no one else is seeing it, how would you feel about just merging this in as is (with the sample env updated)?

robertandremitchell (approved):
lgtm!

m-goggins (approved):
I'm good to merge. I'm not able to reproduce the error after doing some cleanup with docker.

PR #297, opened 2025-01-16
replace `use_cases` with queries from the query table

### LINKED ISSUE #299: Replace hard-coded queries with entries in the query table
## Description:

Since we have custom query building completed, we'll need to replace the hardcoded use cases in the query execution flow with reads to the query table.

## Acceptance Criteria:

Custom queries show up and can be executed in the customize query flow

fzhao99:
# PULL REQUEST

## Summary
Completes the query building work by switching out the hard-coded use cases with the custom queries in the query table!

Had to change some of the ways we're reaching out to the DB on the API side to not use the hard-coded use case constants. Handled these fallbacks to the old param gracefully, but once deployed, we should message that our API has changed and that folks should switch to the new params. Other than handling this deprecation fallback, should have gotten rid of all other hard coded use case constants!

Also updated most of the documentation in the Postman collection, but do let me know if anyone catches something I missed.

## Related Issue
Fixes #299 

## Checklist
- [x] Link to relevant issues
- [ ] Update documentation


robertandremitchell (approved):
looks good to me. a few nits:

* I wonder if we need the Customize Query page? Right now, multi-condition queries have all the valuesets, not split by condition, which can make that page longer and harder to distinguish which valueset belongs to what. I guess I can see the potential of having "unsaved queries" where you could make changes to established queries and see what they look like. But it seems like we would want probably want to just have this go to the /queryBuilding endpoint in the long run rather than maintain both?
* Less relevant, but when you start from a clean build, you have no queries available on the Select query page. I don't think this is the biggest of deals, as if you're building from scratch you probably don't need the demo workflow. We also don't know if we're gonna ship with any default queries anyway. We could consider that if there is no database whenever you go to any endpoint triggering the DB build, but I think that's probably overkill.

fzhao99:
> * I wonder if we need the Customize Query page? 
I think this is coming up in a few places and there's a bigger convo from product / design to nix it in favor of the new workflow!

> * Less relevant, but when you start from a clean build, you have no queries available on the Select query page. I don't think this is the biggest of deals, as if you're building from scratch you probably don't need the demo workflow. We also don't know if we're gonna ship with any default queries anyway. We could consider that if there is no database whenever you go to any endpoint triggering the DB build, but I think that's probably overkill.

ah, good point! theoretically in dev mode the db dump should seed us with the demo queries, but point taken that in the docker setup this won't be the case. to your point, not super worried about it in the short term given folks won't be able to build queries without doing something in the query building flow first (which would necessitate a seed), but probably a UX / navigational challenge we'll need to manage.

PR #298, opened 2025-01-16
refactor fhir calls to use POST

fzhao99:
# PULL REQUEST

## Summary
Refactors our FHIR query layer to 

1. all use POST, since querying by the codes as constructed in the query building flow run into GET URL length limits
1. construct concept type (labs / meds / conditions) filters that we use in the FHIR queries more accurately. Before, the logic was system-based, resulting in some bad queries that returned incorrect amounts of data for certain resources. These are now updated to use the stored information that we have on concept type.
2. Handle the immunization query to still use GET for now since POSTing against their dev server resulted in hostname / IP certificate mismatch errors

Also realized during testing that our e2e's weren't properly selecting / filtering row results for tests that had to do with counting the number of expected resources. For example, the snippet below
```
 await expect(
      page
        .getByRole("table")
        .filter({ hasText: "doxycycline hyclate 100 MG" })
        .getByRole("row"),
    ).toHaveCount(2);
  });
```
gets all tables, filters on the presence of doxycycline hyclate 100 MG on the _tables_, and then counted the rows. These last two steps are swapped / isn't what we wanted. Instead,

```
 await expect(
      page
        .getByRole("table")
        .getByRole("row")
        .filter({ hasText: "doxycycline hyclate 100 MG" })
    ).toHaveCount(2);
  });
```
gets the tables, gets their constituent rows, and then counts the instances of doxycycline hyclate 100 MG. This changed some of the counts (which I verified against a cut of main / the remote HAPI server) that should now be correct.

## Checklist

- [x] Link to relevant issues
- [ ] Update documentation


robertandremitchell (approved):
lgtm!

PR #306, opened 2025-01-22
Dedupe FHIR Resources Parsed by Queries

### LINKED ISSUE #301: Dedupe Resources
## Description:

If we query for a resource in multiple ways, e.g., the resource contains multiple codes, it will be included in the results multiple times. Users have expressed this to be confusing and in order to reduce the amount of data being passed around and displayed to users, we should deduplicate the resources when querying.

## Acceptance Criteria:

- [ ] Resources with the same resource id are not returned in the API or in the UI
- [ ] Resources are deduplicated by resource id in `parseFhirSearch`

m-goggins:
# PULL REQUEST

## Summary

The PR also includes a bit of cleanup on constants that are no longer used (patientOptions and patientType). To dedupe resources added to the `QueryResponse` and then displayed to users, I modified `processFhirResponse` and `parseFhirSearch`.  

The modifications to `processFhirResponse` dedupe resources returned in a given query for a resource type. That is, when we query for Observations using a list of codes, these changes will only add Observations with unique IDs for that query. 

The modifications to `parseFhirSearch` dedupe resources across a whole query, e.g. "My STLT's Chlamydia Query". This is important if we end up pulling back resources of the same type (e.g., Observations) from separately executed queries because the resources would not be deduped by `processFhirResponse` alone. For example, if we query for Observations using a list of codes and then separately query for Observations using the `social-history` class, we could end up returning the same Observation resource in the two queries that would then need to be deduped.

This PR
**Sample Changes using Chlamydia ONE JMC**

Old (returning the same resource multiple times if queried for multiple ways or if the bundle somehow included multiples of the same resource)
image

New (returning only 1 copy of each resource by resource id)
image

## Related Issue

Fixes #301 

## Additional Information
I noticed that we are relying on `FhirResource` that is our own type and I believe we should switch over to using the one provided by `fhir/r4` as it is more robustly typed. As it stands, our `FhirResource` only refers to a handful of resource types, e.g., Observations, Medication Requests, etc. and we would need to manually update the list if we want to include other resource types in future queries, e.g., Practitioner. I suggest we removed the DIBBs `FhirResource` type and switch to using the one that is better typed.

## Checklist

- [x] Link to relevant issues
- [ ] Update documentation

### DIFF
--- a/query-connector/src/app/query-service.ts
+++ b/query-connector/src/app/query-service.ts
@@ -5,7 +5,6 @@ import { Bundle, DomainResource } from "fhir/r4";
 
 import FHIRClient from "./fhir-servers";
 import { DibbsValueSet, isFhirResource, FhirResource } from "./constants";
-
 import { CustomQuery } from "./CustomQuery";
 import { GetPhoneQueryFormats } from "./format-service";
 import { formatValueSetsAsQuerySpec } from "./format-service";
@@ -64,7 +63,7 @@ export type FhirQueryResponse = Awaited<ReturnType<typeof makeFhirQuery>>;
 async function queryEncounters(
   patientId: string,
   fhirClient: FHIRClient,
-  queryResponse: QueryResponse,
+  queryResponse: QueryResponse
 ): Promise<QueryResponse> {
   if (queryResponse.Condition && queryResponse.Condition.length > 0) {
     const conditionId = queryResponse.Condition[0].id;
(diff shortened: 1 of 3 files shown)


query-connector/src/app/query-service.ts
```diff
@@ -220,9 +219,11 @@ async function generalizedQuery(
  */
 export async function parseFhirSearch(
   response: fetch.Response | Array<fetch.Response>,
-  queryResponse: SuperSetQueryResponse = {},
+  queryResponse: Record<string, FhirResource[]> = {},
 ): Promise<QueryResponse> {
-  let resourceArray: SuperSetFhirResource[] = [];
+  let resourceArray: FhirResource[] = [];
+  // let resourceIds: string[] = [];
```
fzhao99, line 225:
```suggestion
```

guessing we meant to delete this!

fzhao99 (approved):
lgtm! and all for using the provided library type over our custom one

robertandremitchell (approved):
looks good to me!

PR #309, opened 2025-01-22
add search to conditions and valueset side panels

fzhao99:
# PULL REQUEST

## Summary

What changes are being proposed? For PRs that include UX/UI changes please make sure to describe their scope and highlight any specific areas that design reviews should focus.

## Related Issue

Fixes #

## Additional Information

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation

### DIFF
--- a/query-connector/src/app/queryBuilding/components/ConceptSelection.tsx
+++ b/query-connector/src/app/queryBuilding/components/ConceptSelection.tsx
@@ -1,12 +1,12 @@
 import React from "react";
 import styles from "../buildFromTemplates/conditionTemplateSelection.module.scss";
-import { Concept } from "@/app/constants";
 import { showToastConfirmation } from "@/app/query/designSystem/toast/Toast";
 import Checkbox from "@/app/query/designSystem/checkbox/Checkbox";
+import { ConceptDisplay } from "./SelectionViewAccordionBody";
 
 type ConceptSelectionProps = {
-  concepts: Concept[];
-  onConceptsChange: (updatedConcepst: Concept[]) => void;
+  concepts: ConceptDisplay[];
+  onConceptsChange: (updatedConcepst: ConceptDisplay[]) => void;
 };
 
 /**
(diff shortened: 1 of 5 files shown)


query-connector/src/app/queryBuilding/components/ConceptSelection.tsx
```diff
@@ -1,12 +1,12 @@
...
-import { Concept } from "@/app/constants";
 import { showToastConfirmation } from "@/app/query/designSystem/toast/Toast";
 import Checkbox from "@/app/query/designSystem/checkbox/Checkbox";
+import { ConceptDisplay } from "./SelectionViewAccordionBody";
 
 type ConceptSelectionProps = {
-  concepts: Concept[];
-  onConceptsChange: (updatedConcepst: Concept[]) => void;
+  concepts: ConceptDisplay[];
+  onConceptsChange: (updatedConcepst: ConceptDisplay[]) => void;
```
robertandremitchell, line 9: can we updated the spelling here? `updatedConcepst`

m-goggins (dismissed):
The search is looking good to me! I think we should update the code to better handle the case where we don't find any matching codes. When I searched "none 100", nothing came up (yay) but I was still able to "add" it (nothing was actually added but the toast popped up). 

Michelle might have different thoughts, but it seems intuitive to me to add a "no matching codes" found message and possibly add something like "would you like to add a new code" once that feature is ready for development.

image

fzhao99:
> for conditions, do we also want to retain the ability to search by category name? I know that is currently possible on the previous page.
This should be the same on the conditions drawer as it is on the other page! Did you find any regressions here that I might have missed?

fzhao99:
> Michelle might have different thoughts, but it seems intuitive to me to add a "no matching codes" found message and possibly add something like "would you like to add a new code" once that feature is ready for development.

Checked with her and added this in the latest commit

https://github.com/user-attachments/assets/8b161b63-e0af-4182-a835-2bde7d8f7a50
https://github.com/user-attachments/assets/96efef03-5817-4685-9c00-2eb7a6a5ea68

robertandremitchell (approved):

query-connector/src/app/queryBuilding/buildFromTemplates/BuildFromTemplates.tsx
```diff
@@ -119,13 +119,11 @@ const BuildFromTemplates: React.FC<BuildFromTemplatesProps> = ({
         return;
       }
       const result = await getSavedQueryById(selectedQuery.queryId);
-
       if (result === undefined) {
         return; // todo: error???
       }
-
       const initialState: NestedQuery = {};
-      const savedQuery = result[0]; // what is this shape, why first index
```
fzhao99, line 128: This is a result of the weird way we're typing `getSavedQueryById` to return an array of results rather than a single one. Will take care of this in a followup PR!

katyasoup (changes requested):
Looks great! Two things I spotted, one that I think shouldn't be blocking and one that maybe is:

1) When searching in the table view, anything that doesn't match my search term gets unchecked:
https://github.com/user-attachments/assets/3b7e8fbe-0845-410a-8481-facea1276180
But, once I delete my search, everything behaves as expected (including persistence of checked/unchecked if I altered that)

~2) On the drawer concept search, it resets when I delete my search; on the Add Condition search, it does not:
https://github.com/user-attachments/assets/c8da75cb-c603-4b1f-9ad6-d7c63194b05d
If I close the drawer and re-open it to Add Condition, it seems to retain my last search, so I have to refresh to get the full condition list - this one I think should be addressed in this PR if possible~
nvm, this was just me not pulling the latest! This looks like it's working now, and shows the "no results found" message as expected 🎉

katyasoup (approved):

fzhao99:
> When searching in the table view, anything that doesn't match my search term gets unchecked:
https://github.com/user-attachments/assets/3b7e8fbe-0845-410a-8481-facea1276180
But, once I delete my search, everything behaves as expected (including persistence of checked/unchecked if I altered that)

Great catch on this point! Went ahead and swapped this out for a log statement since we're taking care of this in an upcoming PR

PR #322, opened 2025-01-28
Creating backend functions for user management

### LINKED ISSUE #272: Create function that checks whether the user already exists in the user management table
## Description:

Now that we are able to [retrieve user info from Keycloak](https://linear.app/skylight-cdc/issue/QUE-126/update-authts-to-return-user-info-for-storage-from-keycloak) when a user authenticates, we need a function that checks whether the user info has been previously stored in the user management table. 

## Acceptance Criteria:

- [ ] A new function is added that accepts a username/email from Keycloak as input and checks whether that username already exists in the user management table and returns a boolean.
- [ ] The new function should run every time a user logs in to the app.

robertandremitchell:
# PULL REQUEST

## Summary

This adds a backend function that runs when you click the sign in button to redirect to keycloak, log you in, then check if the user exists in `user_management` table. if not, then user is added to the table.

Credentials here: https://github.com/CDCgov/dibbs-query-connector/blob/main/query-connector/README.md?plain=1#L88

https://github.com/user-attachments/assets/1c9bf25a-c193-4530-99e5-23830341cbe3

## Related Issue

Fixes #272 

## Additional Information

This also removes a linux/amd64 in the `docker-compose.yml` that @nickclyde identified as something we don't actually need. It seems to be an artifact of the original compose file we had from ecr-viewer.

I also added a docker compose down to npm run dev, because i was getting annoyed at HAPI and other services staying up.

Right now, we are hardcoding qc_role as "super-admin". We will eventually have a default "Standard User" (https://linear.app/skylight-cdc/issue/QUE-128/create-function-that-adds-new-user-info-to-the-user-management-table).

Also curious if at this point we should cut a new `vs_dump.sql` or wait for all the backend work to be complete. personally ambivalent, just means we are adding the `qc-admin` a lot.

I had also started to explore doing this via API, when I was running into issues getting this to work in `auth.ts`. I did not end up needing to use it, but curious if anyone feels strongly we should use API for this.

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation


query-connector/src/app/api/user-management/route.ts
```diff
@@ -0,0 +1,29 @@
+import { NextResponse } from "next/server";
+import { addUserIfNotExists } from "@/app/backend/user-management";
+
+/**
+ * API route for adding a user to the user_management table.
+ * @param req - The incoming request object.
+ * @returns A response indicating success or failure.
+ */
+export async function POST(req: Request) {
+  try {
```
fzhao99, line 10: do we have any auth we can put in front of this endpoint or otherwise do we have any mitigation against unwanted requests? nervous about exposing write functionality into our DB without it

query-connector/src/app/backend/user-management.ts
```diff
@@ -0,0 +1,67 @@
...
+  const userIdentifier = username || email;
+
+  try {
+    console.log("Checking if user exists:", userIdentifier);
+
+    const checkUserQuery = `SELECT username FROM user_management WHERE username = $1;`;
+    const userExists = await dbClient.query(checkUserQuery, [userIdentifier]);
+
+    if (userExists.rows.length > 0) {
+      console.log("User already exists in user_management:", userIdentifier);
```
fzhao99, line 39: Thoughts on swapping this identifier out with the ID so that we're not unnecessarily logging out PII? Not sure what the regulations are for this, but from my perspective better safe than sorry

fzhao99 (commented):
left a couple of security related questions but otherwise looking good!

query-connector/src/app/api/user-management/route.ts
robertandremitchell: makes sense. right now it's unused, so we could either remove it but save the code for a future PR if we did want to implement an API endpoint. It may not be needed, but was one of the things i had looked at when blocked on implementing the user-management script in auth.ts

robertandremitchell: issue created: https://linear.app/skylight-cdc/issue/QUE-158/spike-whether-we-need-api-pathing-for-user-management

query-connector/src/app/backend/user-management.ts
robertandremitchell: updated!

fzhao99 (approved):

PR #324, opened 2025-01-29
fix backend query typing

fzhao99:
Taking care of [this comment](https://github.com/CDCgov/dibbs-query-connector/pull/309/files#r1934063774) which raised an unintuitive way we're typing out a database query.

Our two `get` operations against the query table that query by a unique identifier (name and ID) both used to return an array, even though by definition, there'd only ever be one or zero return values. This made callers require unintuitive `[0]` accesses on responses, which this PR fixes along with some other minor refactoring.


robertandremitchell (approved):
lgtm!

PR #325, opened 2025-01-29
add immunization to query table

fzhao99:
Adds an immunization query like the one we built for the connectathon into our default set of queries

per the conversation documented here: https://skylight-hq.slack.com/archives/C05BEG226RZ/p1738101794406449


nickclyde (approved):
Looks great, thanks Bob!

PR #326, opened 2025-01-29
fix state initialization problems

fzhao99:
Fixes the bug introduced in [this PR](https://github.com/CDCgov/dibbs-query-connector/pull/309) as detailed [here](https://skylight-hq.slack.com/archives/C05BEG226RZ/p1738184273394609)

Is a little late in the day for me so wanted to get the cut of main working first, but will add a test in to catch future instances of this issue first thing tomorrow

https://github.com/user-attachments/assets/a9a8ac57-25a2-45dd-bd59-9731c408f734

### DIFF
--- a/query-connector/src/app/(pages)/queryBuilding/components/SelectionTable.tsx
+++ b/query-connector/src/app/(pages)/queryBuilding/components/SelectionTable.tsx
@@ -7,6 +7,7 @@ import { DibbsConceptType, DibbsValueSet } from "@/app/shared/constants";
 import { ConceptTypeToDibbsVsMap } from "@/app/utils/valueSetTranslation";
 import ConceptTypeAccordionHeader from "./SelectionViewAccordionHeader";
 import MultiAccordion from "@/app/ui/designSystem/MultiAccordion";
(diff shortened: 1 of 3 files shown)


query-connector/src/app/(pages)/queryBuilding/components/SelectionTable.tsx
```diff
@@ -31,11 +32,10 @@ export const ConceptTypeSelectionTable: React.FC<
 > = ({ vsTypeLevelOptions, handleVsTypeLevelUpdate, searchFilter }) => {
   const [expanded, setExpandedGroup] = useState<string>("");
   const [valueSetDisplay, setValueSetDisplay] =
-    useState<ConceptTypeToDibbsVsMap>(vsTypeLevelOptions);
```
fzhao99, line 34: this state was getting initialized badly so that when there were changes to `vsTypeLevelOptions` (ie when the condition tabs were being toggled through) this value wasn't getting updated / the accordions weren't getting regenerated. Setting it to an empty object and then updating the value as a `useEffect` on 62 is correctly flushing the UI with new data

katyasoup (approved):
👍🏻

PR #339, opened 2025-02-04
valueset search part two

### LINKED ISSUE #145: Search from the valueset selection page
* Concept-level search will give the user a deep link into the side panel with the relevant search filter selected
* Style the results according [to Figma](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=536-17291&m=dev), where acting on the hover state will result in the slideover panel just displaying [the selected code](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=506-6896&m=dev)

fzhao99:
# PULL REQUEST

## Summary
Part two of the valueset search effort, which adds the actual search functionality by

- Integrating filtering logic into the valueset table and drawers when search filters are populated 
    - In instances where search is active on the table (ie the "parent search filter"), the result subset is treated as if those are the only concepts that exist. This means bulk toggling / future searches in the drawer ("child search filter") won't populate any results not in the parent subset until the parent  filter is removed. To see this, notice in the below that 0/2 is the selection number when the search filter is on, but when removed, the totals update to 1/3.     
    

    https://github.com/user-attachments/assets/f795357f-75c8-4bea-9382-b404736fecf8

   - Similarly, child drawer searches that don't match the parent filter won't populate concepts  until the parent is removed.     
   https://github.com/user-attachments/assets/bcc3419d-c6cf-4aed-9d37-de0c0407769b

- Adds /refactors some tests to hit the main search flows / handle the edge cases
- Renders results conditionally, with the use of `<Highlighter />` to indicate search matches and tooltips where necessary 
     - Since we're matching on code and concept name in the parent table search, Highlighter is a bit limited on how the stylings appear for the concept matches. Ideally we'd match on the search substring and not the entire match, but since we're only displaying codes, highlighting when there's a match on display name gets tricky. 
     - I elected just to highlight everything in the table view and then provide the substring highlighting on the drawer since that piece of UI has all the info. 
     - Note the difference in the matching for "300" in the below screenshots
     - Screenshot 2025-02-05 at 11 51 55 AM Screenshot 2025-02-05 at 11 51 51 AM
  
## Related Issue
Fixes #145 

## Testing plan
Test that 
- Search on the valueset table matches on codes and highlights (across multiple conditions, valueset types, etc.)
- When the drawer opens, concepts come pre-filtered and subsequent searches only search within that pre-filtered set
- Toggles (individually within the drawer, bulk selection across valueset and accordion work as expected) and populate the correct toasts
- Adding conditions works as expected and populate the correct toasts
- Saving query leaves the state of the valuesets as you left them

## Checklist
- [x] Link to relevant issues

### DIFF
--- a/query-connector/src/app/(pages)/queryBuilding/components/ConceptSelection.tsx
+++ b/query-connector/src/app/(pages)/queryBuilding/components/ConceptSelection.tsx
@@ -1,12 +1,14 @@
 import React from "react";
 import styles from "../buildFromTemplates/conditionTemplateSelection.module.scss";
-import { ConceptDisplay } from "./SelectionViewAccordionBody";
 import { showToastConfirmation } from "@/app/ui/designSystem/toast/Toast";
 import Checkbox from "@/app/ui/designSystem/checkbox/Checkbox";
+import Highlighter from "react-highlight-words";
+import { FilterableConcept } from "./utils";
 
 type ConceptSelectionProps = {
-  concepts: ConceptDisplay[];
-  onConceptsChange: (updatedConcepts: ConceptDisplay[]) => void;
+  concepts: FilterableConcept[];
+  onConceptsChange: (updatedConcepts: FilterableConcept[]) => void;
+  searchFilter?: string[];
 };
 
 /**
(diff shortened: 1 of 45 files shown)


query-connector/src/app/(pages)/queryBuilding/components/ConceptSelection.tsx
```diff
@@ -97,8 +102,22 @@ const ConceptSelection: React.FC<ConceptSelectionProps> = ({
                     onChange={(e) => toggleSingle(index, e.target.checked)}
                   />
                 </td>
-                <td className={styles.conceptCode}>{concept.code}</td>
-                <td>{concept.display}</td>
+                <td className={styles.conceptCode}>
+                  <Highlighter
```
nickclyde, line 106: Super cool component!!

nickclyde (approved):
This looks great! It's a lot of changes that I can't totally follow, but running it locally, it works like a charm!

PR #340, opened 2025-02-05
Initializing users, function to update qc_role

### LINKED ISSUE #273: Create function that adds new user info to the user management table
## Description:

If a user has not logged into the QC before and the user authenticates , the new username/email from Keycloak is added to the table as a new row with their role set to `Standard User`. 

## Acceptance Criteria:

- [ ] If the function to check whether a user has logged in before returns false ( [QUE-127](https://linear.app/skylight-cdc/issue/QUE-127/create-function-that-checks-whether-the-user-already-exists-in-the)), a new row is added to the user management table that adds the user's username/email from Keycloak and sets their role to `Standard User`.

robertandremitchell:
# PULL REQUEST

## Summary

This initializes new users as a standard user role in the user-management table.

One question/concern I had; I had initialize explored creating a `qcRoleSuperAdmin`, `qcRoleAdmin`, and `qcRoleStandardUser` in Keycloak. My initial code included initializing these roles, pushing these roles back into Keycloak with a new master-realm, and having the create/update functions also push those roles back into Keycloak. 

However, the more I thought about it, I was not entirely sure if we wanted or needed to take this step, given that I would think we would a service-agnostic approach to restricting what pages/services a user had access to. Thoughts?

This new function should be the only one you need to pull into the new page to update user roles. As of now, the only user in the `user-management` table is `qc-admin`.

## Related Issue

Fixes #273 

## Additional Information

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation

### DIFF
--- a/query-connector/src/app/backend/user-management.ts
+++ b/query-connector/src/app/backend/user-management.ts
@@ -42,8 +42,8 @@ export async function addUserIfNotExists(userToken: {
 
     console.log("User not found. Proceeding to insert:", id);
 
-    // TODO: Update the role based on the user's group in Keycloak
-    const qc_role = "super-admin";
+    // Default role when adding a new user, which includer Super Admin, Admin, and Standard User.
+    const qc_role = "Standard User";
 
     const insertUserQuery = `
       INSERT INTO user_management (username, qc_role, first_name, last_name)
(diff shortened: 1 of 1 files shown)


query-connector/src/app/backend/user-management.ts
```diff
@@ -65,3 +65,40 @@ export async function addUserIfNotExists(userToken: {
...
+      UPDATE user_management
+      SET qc_role = $1
+      WHERE username = $2
+      RETURNING username, qc_role, first_name, last_name;
+    `;
+
+    const result = await dbClient.query(updateQuery, [newRole, username]);
+
+    if (result.rows.length === 0) {
+      console.error(`User not found: ${username}`);
```
fzhao99, line 94: similar to the previous PR, would suggest we swap this and the username value on 98 (which might contain some sort of PII) with whatever user ID we're getting back from `result`

fzhao99 (commented):
left one comment! agree on your decision to leave our role considerations internal to the app: anything authorization related should probably stay on our side of the app boundary so we can configure them flexibly. In my mind, Keycloak and whatever other IDP should give us authentication only, and any other authorization should be an application-level concern

query-connector/src/app/backend/user-management.ts
robertandremitchell: makes sense! removed

fzhao99 (approved):

nickclyde (approved):
Looks great! I also agree that we should manage authorization completely within our app and leave only authentication to the IdPs.

Issue #206, opened 2024-12-09
Update Routing for formatting/reused components


robertandremitchell:
## Description:

We need to have a single page component across all our existing / future routes that pages can inherit from (that would include things like this ToastContainer, stylings for default page widths, etc.)

## Acceptance Criteria:

PR #365, opened 2025-02-11
Add loading bar at top of page; add Page component to root layout

### LINKED ISSUE #186: Loading bar
## Description:

When the app is loading, display a blue loading progress bar at the top that goes left to right.

## Acceptance Criteria:

katyasoup:
# PULL REQUEST

## Summary
- Adds [next-nprogress-bar](https://github.com/Skyleen77/next-nprogress-bar) to implement a progress/loading bar at the top of the page.
- Reworks the `Page` component to be more re-usable throughout the app (this was partially addressed in the earlier reorg ticket, the rest of the work is done here):
  - `ui/components/page.tsx` contains the `ToastContainer` component, the `SiteAlert` component, and the new `ProgressBar` component, making them accessible to all child pages instead of adding them individually to each page
  - Adds `Page` to the base `layout.tsx` file so that all children are wrapped inside of `Page`
  - Moves the former `page.tsx` file at project root to `(pages)/landingPage` to align with updated structure
  - utilizes `DataContext` to set the current page's `path`, `mode`, or `buildStep` so that it is accessible to other components (namely, the `SiteAlert` component which displays conditional content in certain areas of the app)

https://github.com/user-attachments/assets/06d7a912-67ea-4da3-ac4c-d84e9e14771c

## Related Issue

Fixes #186 #206 

## Additional Information
**Note:** make sure to add DEMO_MODE to your `.env` file to control showing/hiding of the `SiteAlert` component locally

@mikang we have some control over enabling/disabling the loading bar based on specific links - It is triggered anytime the url updates, so it currently _won't_ run for things like `/queryBuilding` where we stay on the same page but display different content. I have toggled it off for the new `/userManagement` page, so that it doesn't fire when switching between the tabs. Anywhere else you'd like it on/off?

## Checklist

- [x] Link to relevant issues
- [ ] Update documentation


mikang:
Thanks @katyasoup ! Is it possible to get it to fire just when the page and data are refreshed? What do you mean by it doesn't load in query building? If you're progressing through steps, it should fire.

Could we please make the height shorter? Around 60% of the current height

katyasoup:
> Thanks @katyasoup ! Is it possible to get it to fire just when the page and data are refreshed? What do you mean by it doesn't load in query building? If you're progressing through steps, it should fire.
> 
> Could we please make the height shorter? Around 60% of the current height

@mikang updated the height to 8px, does that look better?

I also added a delay of 200 ms so it should only fire when a page takes longer than that to load (in my testing this has usually meant the first time I visit a page it will fire, but then going back and forth after that, it is fast enough that it doesn't show).

When I say it doesn't load in query building - it will run on the first time we go to that page (unless it loads in under 200ms), but then progressing through the steps, since the path/url doesn't change, it doesn't trigger the bar. BUT this is where I think the [skeleton screens](https://linear.app/skylight-cdc/issue/QUE-13/add-skeleton-screen-structure-to-loadingview) ticket would fill in the gap (these are the pages where we use the Loading overlay with the spinner currently). So, going from condition selection to value set selection would not show the loading bar, but you'd still get visual feedback from the skeleton screen while waiting for everything to load. Does that sound ok?

query-connector/e2e/query_workflow.spec.ts
```diff
@@ -107,11 +111,14 @@ test.describe("querying with the Query Connector", () => {
...
-    await expect(alert2).toBeVisible();
-    await expect(alert2).toHaveText(
-      `${CONTACT_US_DISCLAIMER_TEXT} ${CONTACT_US_DISCLAIMER_EMAIL}`,
-    );
+
+    if (showSiteAlert) {
+      const alert2 = page.locator(".custom-alert");
+      await expect(alert2).toBeVisible();
+      await expect(alert2).toHaveText(
+        `${CONTACT_US_DISCLAIMER_TEXT} ${CONTACT_US_DISCLAIMER_EMAIL}`,
```
fzhao99, line 119: feel free to do this in a follow up since i know getting these tests to pass was a bear, but could we pull these three similar blocks into a common `checkForSiteAlert()` function?

mikang:
I don't see the loading bar because I have fast wifi haha but I'm sure it looks good. Skeleton screen sounds great! I think we're good to go 👍

fzhao99 (approved):

query-connector/src/app/(pages)/landingPage/landpingPage.module.scss
robertandremitchell, line 1: minor typo in the name

robertandremitchell (approved):
lgtm! one minor typo I saw but reorg makes sense to me.

I'd also recommend making the update to the sample `.env` just so we don't miss that piece for new users.

query-connector/e2e/query_workflow.spec.ts
katyasoup: ✅

PR #366, opened 2025-02-11
Added terraform workflow dependency on CD and keycloak

shanice-skylight:
# PULL REQUEST

## Summary

- Github workflow named Terraform Plan & Terraform Apply will wait until CD workflow is completed and Build Keycloak. The workflow accounts for Keycloak's conditional run and will run if it results in skipped or success. 

## Related Issue

Fixes #
- Automates deployment that is triggered by a merged to main

## Additional Information
- To test, I manually ran this [workflow](https://github.com/CDCgov/dibbs-query-connector/actions/runs/13318828808) 
## Checklist

- [x] Link to relevant issues
- [x] Update documentation


.github/workflows/ecs_terraform.yaml
```diff
@@ -28,8 +32,32 @@ env:
...
+        id: check_a_status
+        run: echo "status=success" >> $GITHUB_OUTPUT
+
+      - name: Check Keycloak Status (if it exists)
+        id: check_b_status
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          echo "Checking if Keycloak has ran."
+          response=$(gh run list --workflow "Build Keycloak" --branch $GITHUB_REF_NAME --json status --jq '.[0]')
```
nickclyde, line 52: Nice, pretty cool use of the gh CLI!

.github/workflows/ecs_terraform.yaml
```diff
@@ -28,8 +32,32 @@ env:
...
+  check-dependencies:
+    runs-on: ubuntu-latest
+    outputs:
+      source_a_status: ${{ steps.check_a_status.outputs.status }}
+      source_b_status: ${{ steps.check_b_status.outputs.status }}
+
+    steps:
+      - name: Check CD Status
+        id: check_a_status
+        run: echo "status=success" >> $GITHUB_OUTPUT
```
nickclyde, line 44: Hmm, it seems like this doesn't actually check the status of the CD run, and just assumes success? Shouldn't we be using the gh cli here as well to check the status?

.github/workflows/ecs_terraform.yaml
```diff
@@ -28,8 +32,32 @@ env:
...
+
+      - name: Check Keycloak Status (if it exists)
+        id: check_b_status
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          echo "Checking if Keycloak has ran."
+          response=$(gh run list --workflow "Build Keycloak" --branch $GITHUB_REF_NAME --json status --jq '.[0]')
+          if [ "$response" = "null" ]; then
+            echo "status=skipped" >> $GITHUB_OUTPUT
```
nickclyde, line 54: Does the gh CLI return the status as "skipped" or "failure"? In that case can we just assign `response` to `status`?

.github/workflows/ecs_terraform.yaml
shanice-skylight: oooh, good catch

.github/workflows/ecs_terraform.yaml
shanice-skylight: It returns the statuses skipped and failure, so I updated the logic to let the workflow continue unless the status is failure or cancelled, in which case it exits with a status of 1. Instead of checking for every possible status, it now simply ensures the workflow continues as long as the status isn't failure or cancelled.

shanice-skylight: Here's the successfully Github Action: https://github.com/CDCgov/dibbs-query-connector/actions/runs/13318828808

nickclyde (approved):
Awesome, thanks Shanice!!

PR #368, opened 2025-02-11
jd/integrate user management with data

johanna-skylight:
# PULL REQUEST

## Summary
This PR integrates the user management -> users tab to start loading data from the DB. 
A super admin user should be able to change the role of users with the `role dropdown`. A confirmation toast will appear to let the user know the change was successful.

Additional improvements in this PR are:
- Toast component was moved to the root layout and removed from all other pages.
- Role was added to the app session.
- `WithAuth` component has been implemented which handles hiding content until a user session is detected. 
-  Some helper methods have been added to introduce role checks.

## Related Issue

Partially implements [ticket 129](https://linear.app/skylight-cdc/issue/QUE-129/build-user-management-page-view-for-super-admins)

## Additional Information

I will tackle unit and integration testing in a follow up PR.

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation


johanna-skylight:
@robertandremitchell I am also adding/modifying the `user-management.ts` file while working on the data integration with the pages. I am putting my code available as a draft so you can see the changes I am introducing.

query-connector/src/app/(pages)/userManagement/page.tsx
```diff
@@ -4,22 +4,128 @@ import { useContext, useEffect, useState } from "react";
...
+import { useSession } from "next-auth/react";
 
 /**
  * User section in the user management page
  * @returns Users table
  */
 const UserManagement: React.FC = () => {
-  const [users, setUsers] = useState<User[]>([]);
+  const { data: session } = useSession();
+  console.log("user management session:", session);
```
johanna-skylight, line 25: @katyasoup after this PR is merged the user's role will be available in the session object

johanna-skylight:
I am still working on this PR but I would like input on some naming and folders structure locations. @fzhao99 I will tag you in the files if that is ok with you!

query-connector/src/app/models/entities/user-management.ts
```diff
@@ -0,0 +1,24 @@
...
+    | RoleTypeValues.Standard;
+  userGroups?: UserGroup[];
+}
+
+export interface UserGroup {
+  id: string;
+  name: string;
+  memberSize: number;
+  querySize: number;
+}
```
johanna-skylight, line 24: @fzhao99 introduces entities folder for types that map directly to our tables.

query-connector/src/app/ui/components/header/header.tsx
```diff
@@ -186,12 +187,41 @@ const HeaderComponent: React.FC<{ authDisabled: boolean }> = ({
...
+];
+pagesRoleAccess[PAGES.FHIR_SERVERS] = [
+  RoleTypeValues.SuperAdmin,
+  RoleTypeValues.Admin,
+];
+pagesRoleAccess[PAGES.USER_MANAGEMENT] = [RoleTypeValues.SuperAdmin];
+pagesRoleAccess[PAGES.GROUP_MANAGEMENT] = [
+  RoleTypeValues.SuperAdmin,
+  RoleTypeValues.Admin,
 ];
```
johanna-skylight, line 225: @fzhao99 I am trying to get away from typing the urls directly and to centralize the pages and their access... I am wondering if all these should not even be in the `header component`.

query-connector/src/app/utils/configChecks.ts
```diff
@@ -0,0 +1,15 @@
...
+  return process.env.DEMO_MODE === "true";
+}
+
+/**
+ * Checks if the property AUTH_DISABLED is true
+ * @returns true if auth is disabled for the application
+ */
+export function isAuthDisabled(): boolean {
+  return process.env.AUTH_DISABLED === "true";
+}
```
johanna-skylight, line 15: I am not convinced the `configChecks.ts` is the right name for this file.

fzhao99: could we just stick these two methods in `auth.ts`?

query-connector/src/app/models/responses/collections.ts
```diff
@@ -0,0 +1,12 @@
...
+  totalItems: number;
+}
+
+export interface QCPagedResponse<T> extends QCResponse<T> {
+  pageIndex: number;
+  pageSize: number;
+  totalPages: number;
+  prevPage: number;
+  nextPage: number;
+}
```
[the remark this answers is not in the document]
fzhao99: Can see the value of having the paged response struct, but could you say a bit more about why you  prefer wrapping the result in `QCResponse` rather than just returning an array of the objects being manipulated? That would give us direct access to the underlying data in the array over needing to chain off of `items`. The length of the array would similarly give us back the value of `totalItems` while allowing us to do further array operations like filter, map, forEach, etc.

Guess it's not immediately obvious to me what the benefit of having an extra wrapper here is, but am all ears if you have other ideas!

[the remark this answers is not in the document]
fzhao99: Return arrays would also allow us to return empty arrays over nulls in the empty case, so any UI that takes arrays and iterates over them wouldn't have to explicit null checks. Fully admit this one is more of a style / preference thing, but just working with the underlying arrays rather than checking for nulls might save us code in the long run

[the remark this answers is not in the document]
johanna-skylight: This is to start implementing DALs (data access layer) and DTOs (data transfer objects). Is an approach that I have normally follow in other projects but the idea is that you only expose the necessary information to each of the layers of your application. The frontend does not have to see the entire complex object that comes back from the DB query.

There is a small mention in this link https://nextjs.org/docs/app/building-your-application/authentication#using-data-transfer-objects-dto but I am sure there are more detailed articles around.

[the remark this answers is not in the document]
fzhao99: ah, so for my understanding: the idea here eventually is to simplify down the user object further so that the frontend only has the values that we need, compared to the user object coming from the db?

if so, makes sense. thanks for explaining!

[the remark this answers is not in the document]
johanna-skylight: I am still returning the User as the main object but in the context of a list of users for the frontend not in the context of a query response anymore. So the method that retrieves the data has a signature of `export async function getUsers(): Promise<QCResponse<User>>` so the frontend knows it will received QCResponse<User> which is basically an array of User. 

So this allows me to use types more strongly in the code and not to work with the query result type at all in the frontend code.

I am genuinely what all devs will think of this patten because I personally see the benefit but maybe others won't XD.

query-connector/src/app/utils/configChecks.ts
johanna-skylight: This gave me an idea. I created an auth.ts file in the utils folder!

query-connector/src/app/utils/auth.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+export function isAuthDisabled(): boolean {
+  return process.env.AUTH_DISABLED === "true";
+}
+
+/**
+ * Retrieves the logged user. Will only contain the information retrieved from the IDP
+ * @returns User object in the session or undefined if there is no active session
+ */
+export async function getLoggedUser(): Promise<User | undefined> {
```
fzhao99, line 26: is this meant to be `getLoggedInUser()`?

query-connector/src/app/models/responses/collections.ts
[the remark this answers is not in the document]
fzhao99: happy to try this, especially if it's a pattern you've used before. Think at the end of the day we're still dealing with lists of objects: the quibble is whether the extra wrapping of `.items` and `QCResponse` adds value or not. Will keep an eye on it once we have more instances of the abstraction 👀

fzhao99:
Great work on this Johanna! Think the way we've architected the auth components is really smart. Two things:

-  I'm noticing is that session changes seem to require a log out in order to refresh on the frontend. If I have a Super Admin that switches themselves to a standard user (setting aside whether we should allow users to self-demote), it seems that they can still see the super-admin protected pages up until a log out. Is there an additional step for the session update for us to flush wherever we're keeping the session role? Seems like we'd want those protections to kick in immediately after they get changed in the DB.

https://github.com/user-attachments/assets/7f66618f-2002-49e3-88ab-aa88125ee96d

- Relatedly, logging out seems to only destroy the session on the frontend w/o destroying the cookie that we're checking to kick the user to Keycloak for username / password. Not something that I think we need to handle in this PR, but wondering if log out should [also destroy the session within Keycloak](https://stackoverflow.com/questions/71872587/logout-from-next-auth-with-keycloak-provider-not-works), which I don't think comes out of the box from Next Auth  

https://github.com/user-attachments/assets/c53e3e8c-9d70-400e-8bd7-9775e1a59078

johanna-skylight:
@fzhao99 Thanks for catching the issue of the role access not getting updated!

I spent sometime trying to understand how next-auth is creating/modifying the session object for the client side and I have added improvements across the app.

1- I am retrieving the session object in the root layout on the server to initialize the retrieval before first load.
2- I added an extra check in the `WithAuth` component so it differentiates between a session that is still being retrieved vs a session that does not exists.
3- I modified the conditional that shows the `sign in` button on the header so it shows only after a session was not found  and to not display while the session is being retrieved.
4- I fixed the checks in the DAL. They were properly placed but I was not awaiting the checks so the code was basically ignoring them. Hence you were able to keep modifying the user even after the role was downgraded.

I think the point you made about the logout and how it should remove the cookies is valid! we should look into it. I also think we should add logic in the frontend to logout the user automatically and also start checking for timeout sessions??? I don't think anything regarding sessions lifespan is coded yet??? I might be wrong.

Thanks for the review! I think I have fixed your immediate concerns.

query-connector/src/app/ui/components/withAuth/WithAuth.tsx
```diff
@@ -29,7 +29,12 @@ const WithAuth: React.FC<React.PropsWithChildren> = ({ children }) => {
     !access.includes(session?.user?.role as RoleTypeValues)
   ) {
     redirect("/unauthorized");
+  } else if (session === null) {
```
fzhao99, line 32: think we might be able to use [the `status` variable](https://next-auth.js.org/getting-started/client#example) (which correspond to the different values of session here) that next-auth gives us in order to make these checks more readable

query-connector/src/app/backend/user-management.ts
```diff
@@ -116,14 +117,14 @@ export async function updateUserRole(
  * @returns List of users registered in qc
  */
 export async function getUsers(): Promise<QCResponse<User>> {
-  if (!superAdminAccessCheck()) {
+  if (!(await superAdminAccessCheck())) {
```
fzhao99, line 120: It looks like these fixes fixed the access update server side, but client side, think we still need to flush the session accordingly so that any pages that the new user shouldn't see are protected. I would expect that after the update, refreshing this page / trying to navigate to this page with the new permissions should result in an unauthorized error. 

https://github.com/user-attachments/assets/f220aab5-769a-4567-a0fc-c5f7b14bce91

Think either the [update](https://next-auth.js.org/getting-started/client#updating-the-session) or the [refetch](https://next-auth.js.org/getting-started/client#refetching-the-session) methods would probably be what we'd want to use.

johanna-skylight: You are right. And we are kinda on the same page. When I mentioned that we need to implement a way to automatically log out a user and also start checking for timeout sessions, I was basically referring about the strategy explained in the refetch method section. Once the SPA is fully running on the client, the session that was retrieved on first load stays in the browser memory. The client is always the one that starts the conversation with the server so we would need to implement a method for the client to trigger a session refresher on some sort of interval (similar to the example in the fetch session section in the doc you shared). 

Normally what happens is that a component gets created and integrated at the root of the app. That `session timeout` component handles that refetch and shows a modal to the user asking if they want to `refresh the session` but that component is out of the scope for this ticket. 

The update strategy in this case will not work because the `Super admin` can change the role/access level to any user. So in the case of changing its own user role the update strategy would work to resync the session role. But if the `Super admin` changes the role of another user, updating the session of the super admin does not work on updating the session role of that other user, as that other user's app and memory is running on other machine and will be unaware of what has happened. At that point the user with the downgraded permissions will not be able to perform any action because the DAL checks will kick in and if the user refreshes the page then the new session role should be visible. A `session timeout` component will help with this scenario.

I am unaware if nextjs or next-auth allows for easy implementing other protocols that allow the server to push updates to clients (like websockets) but IMO this would be too fancy to implement for this app. 

I think session handling in general should be discussed with all devs and proper tickets created for whatever is agreed as a solution. 

CC: @nickclyde

query-connector/src/app/ui/components/withAuth/WithAuth.tsx
johanna-skylight: Thanks! I will work on this update

fzhao99 (approved):

query-connector/src/app/backend/user-management.ts
nickclyde: Agreed, there is a lot to consider here, and we should def discuss w/ the whole eng team. My initial thoughts here are:

1. IMO we should just not allow Super admins to demote themselves
2. We may even want to prevent promotions to Super Admin via the UI, and only allow assigning super admins via direct database modifications. 
3. When a user's role is changed, their session should be invalidated.
4. Whenever anybody tries to take an action like make a query, build a query, or access user management, we should make the app double check whether their session is still valid. If not, remove the session cookie and redirect to the homepage, forcing the user to login again

Of course this all isn't necessarily in the scope of this PR, and we can continue to adjust as we move forward!

query-connector/src/app/(pages)/userManagement/page.tsx
```diff
@@ -3,23 +3,125 @@
...
+
+  /**
+   * DOM helpers
+   */
+
+  function renderUserRows(users: User[] | null): React.ReactNode {
+    if (users == null) {
+      return (
+        <tr>
+          <td colSpan={3}>Loading...</td>
```
nickclyde, line 72: Nice!!

query-connector/src/app/backend/user-management.ts
nickclyde: In order to be able to invalidate sessions though, we may need to switch to database sessions:
https://authjs.dev/concepts/session-strategies

nickclyde (approved):
Awesome work Johanna! Works well when I run locally. We definitely need to think more as a whole team around session management, but this is a great start while we think through that.

PR #381, opened 2025-02-18
Super admin testing

johanna-skylight:
# PULL REQUEST

## Summary
This PR implements unit testing for the user management page and reusable components introduced with this work.

## Related Issue

Fixes [QUE -129](https://linear.app/skylight-cdc/issue/QUE-129/build-user-management-page-view-for-super-admins)

## Additional Information
Because of the query connector application setup in which docker environment variables are use instead of`.env` files we cannot use runtime environment variables (aka variables prefixed with NEXT_PUBLIC). Some components do need to access the environment variables defined in the different docker and docker compose files at runtime, so I had to implement a workaround in which the runtime variables are fetch on server side and passed in the first rendered html (see changes in the `root layout` which is a server component).

## Checklist
- [ ] Link to relevant issues
- [ ] Update documentation


query-connector/src/app/backend/user-management.ts
```diff
@@ -43,10 +43,17 @@ export async function addUserIfNotExists(userToken: {
...
+    let qc_role = RoleTypeValues.Standard;
     console.log("User not found. Proceeding to insert.");
 
-    // Default role when adding a new user, which includes Super Admin, Admin, and Standard User.
-    const qc_role = "Standard User";
+    // First registered user is set as Super Admin
+    const queryUserRecordCount = `SELECT COUNT(*) FROM users`;
+    const userCount = await dbClient.query(queryUserRecordCount);
+
+    if (userCount?.rows?.[0]?.count === "0") {
```
fzhao99, line 54: could we add another guard here for whether `isAuthDisabled` is true, or another check for local dev? A little worried that this might get accidentally left in

johanna-skylight: will do!

johanna-skylight: I have added the conditional

query-connector/src/app/tests/integration/user_management.test.ts
```diff
@@ -48,12 +56,15 @@ describe("User Management Integration Tests", () => {
   /**
    * Tests adding a new user if they do not already exist.
    */
-  test("should add a user if they do not exist", async () => {
-    const result = await addUserIfNotExists(TEST_USER);
+  test.each([
+    { user: TEST_SUPER_USER, role: RoleTypeValues.SuperAdmin },
+    { user: TEST_USER, role: RoleTypeValues.Standard },
+  ])("should add a user if they do not exist", async ({ user, role }) => {
+    const result = await addUserIfNotExists(user);
```
fzhao99, line 63: feel free to do this in a followup!

it looks like this function returns the user if they do exist. from where i'm reading, we might want to assert that they don't exist first if we're trying to test the state where they get added if they don't exist. in that case, may also want to do some cleanup after to delete the newly added user.

query-connector/src/app/utils/auth.ts
```diff
@@ -13,6 +13,19 @@ export function isDemoMode(): boolean {
...
+): boolean {
+  return runtimeConfig?.AUTH_DISABLED === "true";
+}
+
+/**
+ * Checks if the property AUTH_DISABLED is true
+ * USE ONLY ON SERVER CODE
  * @returns true if auth is disabled for the application
  */
 export function isAuthDisabled(): boolean {
```
fzhao99, line 31:
```suggestion
export function isAuthDisabledServerCheck(): boolean {
```

fzhao99 (approved):
great work powering through the env issues Johanna!

query-connector/src/app/utils/auth.ts
johanna-skylight: Done!

query-connector/src/app/utils/auth.ts
```diff
@@ -13,6 +13,19 @@ export function isDemoMode(): boolean {
 
 /**
  * Checks if the property AUTH_DISABLED is true
+ * USE ONLY ON FRONTEND COMPONENTS
+ * @param runtimeConfig - object that contains all runtime variables
+ * @returns true if auth is disabled for the application
+ */
+export function isAuthDisabledAtRuntime(
```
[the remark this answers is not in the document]
johanna-skylight: done!

PR #389, opened 2025-02-21
Initialize backend add/delete for usergroup_to_users

### LINKED ISSUE #372: Create a backend function to INSERT/DELETE users in usergroup
## Description:

The usergroup to users conjunction table is necessary to manage the users.id that are part of a particular user group. 

The easiest way to do this will be to query the users table for all ids/names to display in the drawer with a boolean to indicate if they are part of a usergroup.

If they are checked, that user_id/usergroup_id combination should have a row inserted into the usergroup_to_users table.

If they are unchecked, that user_id/usergroup_id combination should have a row deleted from usergroup_to_users table, if it exists.

This should be done via a Save Changes buttn, to avoid doing multiple insert/delete states.

## Acceptance Criteria:

robertandremitchell:
# PULL REQUEST

## Summary

This adds the functions to both display usergroup_to_users in the Drawer and also to add/delete in the backend based on their is_member boolean.

The logic to do this for queries should pretty much be 1:1, so will be curious if there are aspects of this work you all think might work better in the frontend.

## Related Issue

Fixes #372 

## Additional Information

I'm also curious if anyone feels strongly about keeping all this work in `user-management.ts`. I had originally had it there, but with the very similar work we are about to do on usergroup_to_queries, that program would be around ~800 lines, so could be a little trickier to navigate.

On the other hand, user management work would now require two imports, fwiw. I don't feel strongly either way.

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation


query-connector/src/app/backend/usergroup-management.ts
```diff
@@ -0,0 +1,155 @@
...
+  const existingUsersResult = await dbClient.query(existingUsersQuery, [
+    userIds,
+  ]);
+  const existingUserIds = new Set(
+    existingUsersResult.rows.map((row) => row.id),
+  );
+  const validUserIds = userIds.filter((id) => existingUserIds.has(id));
+
+  // If no valid user IDs are found, return an empty array
+  if (!validUserIds.length) {
```
fzhao99, line 76: think we can just return the result of the filter without this check, since we'll just have an empty array if filter doesn't get a hit on any of the elements

query-connector/src/app/backend/usergroup-management.ts
```diff
@@ -0,0 +1,155 @@
...
+    return [];
+  }
+
+  const values = validUserIds
+    .map((userId) => `('${userId}_${groupId}', '${userId}', '${groupId}')`)
+    .join(",");
+  const query = `
+    INSERT INTO usergroup_to_users (id, user_id, usergroup_id)
+    VALUES ${values}
```
fzhao99, line 85: we should probably sanitize these values in this insert instead of interpolating the strings in the event the values object somehow gets injected 

example [here](https://node-postgres.com/features/queries#query-config-object)

query-connector/src/app/backend/usergroup-management.ts
fzhao99, line 76: nit: could we check for `.length === 0` here?

query-connector/src/app/tests/integration/usergroup-management.test.ts
```diff
@@ -0,0 +1,176 @@
...
+  });
+
+  /**
+   * Tests retrieving user group memberships.
+   */
+  test("should retrieve user group memberships", async () => {
+    const result = await getUsersWithGroupStatus(TEST_GROUP_ID);
+
+    expect(Array.isArray(result)).toBe(true);
+    if (result.length) {
```
fzhao99, line 65: could we swap this out with an explicit check for `result[0]` being what we'd expect it to? As is, if result returns an empty array, think this test would pass even when I think (but correct if I'm mistaken!) it should fail

fzhao99:
left a few notes! think splitting these out to multiple files makes sense.

johanna-skylight (commented):
Good work @robertandremitchell . I have left some comments/questions thanks!

query-connector/src/app/backend/usergroup-management.ts
```diff
@@ -0,0 +1,155 @@
...
+ * @returns A list of user group memberships for the specified group.
+ */
+export async function getUsersWithGroupStatus(
+  groupId: string,
+): Promise<UserGroupMembership[]> {
+  if (!(await adminAccessCheck())) {
+    throw new Error("Unauthorized");
+  }
+
+  const groupCheckQuery = `SELECT id FROM usergroup WHERE id = $1;`;
```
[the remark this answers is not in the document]
robertandremitchell: I was thinking through the edgiest of edge cases where users are being generated/deleted at the same time someone is trying to create a user group. 

I don't think we will be that will be likely, but yeah, happy to remove.

query-connector/src/app/backend/usergroup-management.ts
```diff
@@ -0,0 +1,155 @@
...
+  if (!userIds.length) return [];
+
+  const existingUsersQuery = `SELECT id FROM users WHERE id = ANY($1);`;
+  const existingUsersResult = await dbClient.query(existingUsersQuery, [
+    userIds,
+  ]);
+  const existingUserIds = new Set(
+    existingUsersResult.rows.map((row) => row.id),
+  );
+  const validUserIds = userIds.filter((id) => existingUserIds.has(id));
```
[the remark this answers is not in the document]
robertandremitchell: same note as above; just the very edgy possibility someone is deleted in the midst of a usergroup update.

query-connector/src/app/models/entities/user-management.ts
```diff
@@ -22,3 +22,10 @@ export interface UserGroup {
   memberSize: number;
   querySize: number;
 }
+
+export interface UserGroupMembership {
+  id: string;
+  user: User;
+  usergroup_id: string;
+  is_member: boolean;
```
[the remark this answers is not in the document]
robertandremitchell: so you think we should use `UserGroupMembership` as an optional group array on `User`? I think that makes a lot of sense, I can update.

query-connector/src/app/backend/usergroup-management.ts
robertandremitchell: makes sense, switching to query config, lmk what y'all think!

query-connector/src/app/tests/integration/usergroup-management.test.ts
```diff
@@ -0,0 +1,176 @@
...
+
+const dbClient = getDbClient();
+
+jest.mock("@/app/utils/auth", () => ({
+  adminAccessCheck: jest.fn(() => Promise.resolve(true)),
+}));
+
+const TEST_GROUP_ID = "00000000-0000-0000-0000-000000000000";
+const TEST_USER_1 = "00000000-0000-0000-0000-000000000001";
+const TEST_USER_2 = "00000000-0000-0000-0000-000000000002";
```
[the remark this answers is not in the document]
robertandremitchell: added!

query-connector/src/app/backend/usergroup-management.ts
robertandremitchell: updated!

query-connector/src/app/backend/user-management.ts
```diff
@@ -316,6 +316,14 @@ export async function deleteUserGroup(id: string): Promise<UserGroup | string> {
...
 
   try {
+    await dbClient.query(
+      "DELETE FROM usergroup_to_users WHERE usergroup_id = $1",
+      [id],
+    );
+    await dbClient.query(
+      "DELETE FROM usergroup_to_query WHERE usergroup_id = $1",
+      [id],
+    );
```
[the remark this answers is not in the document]
robertandremitchell: I think the main reason is just transparency in the deletion order and to make explicit the relation between the tables. I'm happy to switch if we think that makes more sense here, I mostly just prefer deleting sequentially/independently for the sake of clarity.

query-connector/src/app/backend/usergroup-management.ts
robertandremitchell: https://docs.google.com/document/d/1O0Y-t--QmvqcZGVXCppqU-_2OAR0qiTjiRMar5wkuSk/edit?tab=t.0

Adding as a discussion point for tomorrow; unfortunately I can't be there but the large topic might be good to chat about with the full team.

query-connector/src/app/tests/integration/usergroup-management.test.ts
```diff
@@ -0,0 +1,194 @@
...
+      expect(result[0]).toHaveProperty("id");
+      expect(result[0]).toHaveProperty("username");
+      expect(result[0]).toHaveProperty("userGroupMemberships");
+
+      const membership = result[0].userGroupMemberships?.find(
+        (m) => m.usergroup_id === TEST_GROUP_ID,
+      );
+      expect(membership).toBeDefined();
+      expect(membership?.is_member).toBeDefined();
+    }
```
fzhao99, line 77:
```suggestion
expect(result[0]).toHaveProperty("id");
expect(result[0]).toHaveProperty("username");
expect(result[0]).toHaveProperty("userGroupMemberships");

const membership = result[0].userGroupMemberships?.find(
  (m) => m.usergroup_id === TEST_GROUP_ID,
);
expect(membership).toBeDefined();
expect(membership?.is_member).toBeDefined();
```

Think we're better off not doing this check, since in the event that `result[0]` is malformed, the if actually prevents us from catching it

query-connector/src/app/backend/usergroup-management.ts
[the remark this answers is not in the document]
robertandremitchell: NOTE to self: Remove the check for get since it should not be impacted.

query-connector/src/app/backend/usergroup-management.ts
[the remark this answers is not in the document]
robertandremitchell: Resolved: remove the extra check to rely on FK.

query-connector/src/app/models/entities/user-management.ts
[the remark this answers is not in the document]
robertandremitchell: Resolved: update getUsers() with user group information

query-connector/src/app/backend/usergroup-management.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  if (validUserIds.length === 0) {
+    return [];
+  }
+
+  const insertQuery = {
+    text: `
+      INSERT INTO usergroup_to_users (id, user_id, usergroup_id)
+      VALUES ${validUserIds.map((_, i) => `($${i * 3 + 1}, $${i * 3 + 2}, $${i * 3 + 3})`).join(",")}
```
nickclyde, line 91: I'm a little confused by the syntax here - are you performing arithmetic on the IDs? Is this just to ensure uniqueness?

robertandremitchell: yeah, this was something I found online for how to try to do this, but I don't think it should be necessary. simplifying.

robertandremitchell:
Found a nice library for doing dynamic inputting without risking SQL injection: https://www.npmjs.com/package/pg-format

fzhao99 (approved):

Issue #386, opened 2025-02-20
SPIKE: Write design doc for audit_logs table


nickclyde:
## Description:

Think through how to create a table for storing audit logs.

## Acceptance Criteria:

fzhao99:
Should include:

* DB design options / possibilities and how they link back to the ONC docs (should make sure we have all the pieces in the design to account for the ONC regs)
* Potential tools / libraries we could use
* The pseudocode for how we might build out the developer API
  * Could use Typescript decorators or something similar to centralize the logging logic

PR #402, opened 2025-03-04
JD/restrict page access by role

johanna-skylight:
# PULL REQUEST

## Summary

This PR implements [QUE-131 Restrict access to pages by user roles](https://linear.app/skylight-cdc/issue/QUE-131/restrict-access-to-pages-by-user-roles). This code introduces checks to the `Query`, `My Queries` and `FHIR Servers` pages. Frontend checks and handling is added by wrapping the pages with the `WithAuth` component following pattern mentioned in [auth.js documentation](https://authjs.dev/guides/role-based-access-control?framework=next-js#using-the-role). Backend checks and handling is added by adding role checks to the methods that retrieve the data for the pages, following the DAL pattern described in this [Next.js doc](https://nextjs.org/blog/security-nextjs-server-components-actions#data-access-layer). The Query Connector application had a folder structure that partially implemented the DAL pattern. In the code the folder that contains the methods that retrieve the data from the DB and populate the frontend pages are placed in the `backend` folder, that folder is being treated as the DAL.

_Other code restructuring introduced_

Some methods that connect to the DB and were used to retrieve data to populated the `Query`, `My Queries` and `FHIR Servers` pages were moved out from the file `query-connector/src/app/shared/database-service.ts` and placed in the files `fhir-servers.ts` and `query-building.ts` in the `backend` folder. This is related to the DAL pattern mentioned above.

The types that represented the entities in our DB schema have been migrated from the `query-connector/src/app/shared/constants.ts` to the files `concepts.ts`, `fhir-servers.ts`, `query.ts`, `valuesets.ts` in the `models/entities` folder. The file `user-management.ts` has been renamed to `users.ts`. All the files now match their table name in the DB.

I also added mocks for the next-auth imports to improve the mocking of user session during integration testing.

## Related Issue
Fixes [QUE-131 Restrict access to pages by user roles](https://linear.app/skylight-cdc/issue/QUE-131/restrict-access-to-pages-by-user-roles)

## Additional Information
A lot of the files changes are update to imports.

Session related items that are out of scope for this work:
- Session sync between frontend and backend.
- Session refresh and auto log out.
- Invalidating a session.

## Checklist

- [ ] Descriptive Pull Request title
- [ ] Link to relevant issues
- [ ] Provide necessary context for design reviewers
- [ ] Ensure test coverage is above agreed upon threshold
- [ ] Update documentation

PR #405, opened 2025-03-04
PROOF OF CONCEPT - Audit log prototype

fzhao99:
# PULL REQUEST

## Summary
Exploratory PR to use [Typescript decorators](https://www.typescriptlang.org/docs/handbook/decorators.html#method-decorators) for the audit logging effort. 

Exposes an `@auditable()` decorator that 
- Appropriately handles sync and async functions
- Validates we can make a round trip to the db with the appropriate with result values

## Related Issue
Part of #386 

## Additional Information

One major con:
**Decorators only work for classes** Think in most cases though, the idea of a "service class" to interact with something outside the next app (ie a FHIR client, an IDP, the database that would necessitate a auditable event) works quite well / is an architectural pattern that we'd might want to adopt anyways.

This might require a bit of reorganization for any DB methods / future methods that need auditing, but I think the centralization / syntatic sugar of annotating the underlying method once and forgetting any auditing concerns from there is quite nice. 

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation

### DIFF
--- a/query-connector/src/app/shared/CustomQuery.ts
+++ b/query-connector/src/app/shared/CustomQuery.ts
@@ -1,4 +1,5 @@
 import { QueryDataColumn } from "../(pages)/queryBuilding/utils";
+import { auditable } from "./decorators";
 
 /**
  * A Data Class designed to store and manipulate various code values used
@@ -201,6 +202,7 @@ export class CustomQuery {
     };
   }
 
+  @auditable()
   compileAllPostRequests() {
     return Object.values(this.fhirResourceQueries)
       .map((q) => {
--- a/query-connector/src/app/shared/decorators.ts
+++ b/query-connector/src/app/shared/decorators.ts
@@ -1,0 +1,55 @@
+import { getDbClient } from "../backend/dbClient";
+
+export function auditable(async = false) {
+  return function (
+    target: Object,
+    key: string,
+    descriptor: PropertyDescriptor,
+  ) {
+    const dbConnection = getDbClient();
+    const method = descriptor.value;
+
+    const logAndReturnResult = (resultToLog: unknown, args?: unknown[]) => {
+      // would replace the below round trip to the DB with the actual write to
+      // the table
+      const query = "SELECT * FROM conditions";
+      const result = dbConnection.query(query);
+
+      result.then((v) => {
+        console.log(v.rows[0]); // in case we need to do anything with the db response
+        console.log("To add to audit log: ", resultToLog); // can write the result
+        if (args) {
+          console.log("Args to add to the audit log log: ", args); // or the args
+        }
+      });
+
+      return resultToLog;
+    };
+    // need to do this check so that async vs sync returns are properly typed
+    descriptor.value = async
+      ? async function (this: unknown, ...args: unknown[]) {
+          try {
+            const resultToLog = await method.apply(this, args);
+            logAndReturnResult("", args);
+
+            return resultToLog;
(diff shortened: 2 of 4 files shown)


johanna-skylight:
@fzhao99 do you know if the annotation can be used in static methods in the class?

I am thinking that for example we could just wrap all the methods in backend/user-management.ts in a class and then migrate the method calls to use the class prefix `UserManagement.getUsers()` so we would not need to even create a new object per page/file... etc. Just wondering 🤔

query-connector/src/app/(pages)/fhirServers/page.tsx
```diff
@@ -2,11 +2,11 @@
 
 import { Icon, Label, TextInput } from "@trussworks/react-uswds";
 import {
-  deleteFhirServer,
-  getFhirServersList,
   insertFhirServer,
   updateFhirServer,
-} from "./../../backend/fhir-servers";
+  deleteFhirServer,
+  getFhirServerConfigs,
```
robertandremitchell, line 8: is this a leftover from a merge? I don't have a strongly held opinion between Configs v List, just wasn't sure if it was intentional.

robertandremitchell (approved):
lgtm, one minor comment but definitely not blocking

query-connector/src/app/(pages)/fhirServers/page.tsx
fzhao99: Think this was a rename from [Johanna's user management PR](https://github.com/CDCgov/dibbs-query-connector/pull/402/files#diff-65478080d3b962220420f4f391e0415589b0da2b9280c301b467ba704b8354c8R9) that needed updating

johanna-skylight: I created a wrapping method for the `getFhirServerConfigs ` that checks the role before returning data to the FHIR server page. But other pages retrieve the list of servers like the `query` page. So the method used in query does not check roles and returns the fhir servers as usual.... the method used in fhir servers page makes a role check and throws errors accordingly. Hope that clarifies my PR changes!.

fzhao99: Ah ok! I'll go back and fix that then. I think I dropped the extra admin check in the moving of code around

nickclyde (approved):
Awesome work, Bob!!

query-connector/src/app/backend/dbServices/fhir-servers.ts
```diff
@@ -0,0 +1,295 @@
...
+class FhirServerConfigService {
+  private static dbClient: Pool = getDbClient();
+  private static cachedFhirServerConfigs: FhirServerConfig[] | null = null;
+
+  /**
+   * Fetches the configuration for a FHIR server from the database.
+   * @param forceRefresh - Whether to flush the config cache
+   * @returns The configuration for the FHIR server.
+   */
+  static async getFhirServerConfigs(forceRefresh = false) {
```
johanna-skylight, line 17: Can you make sure [this code](https://github.com/CDCgov/dibbs-query-connector/blob/0d7eaceb388434284d3b646ffd73848631ad5c44/query-connector/src/app/backend/fhir-servers.ts#L289) gets migrated?. These are the methods that I was talking about where one needs to perform role checks and the other does not. 

The query page should be consuming getFhirServerConfigs and the fhir servers page getFhirServersList. Right now in this branch the query page has runtime errors because of the role access checks.

PR #414, opened 2025-03-10
Audit Log Frontend Page

### LINKED ISSUE #353: Frontend: Build Audit Log page
## Description:

This ticket will focus on building the `/audit-log` endpoint and myriad frontend components. Subsequent tickets will focus on connecting this to the backend audit work.

## Acceptance Criteria:

robertandremitchell:
# PULL REQUEST

## Summary

Initializing audit log front endpoint and building page skeleton.

@mikang , please let me know what design updates ought to be made. As part of a separate convo, it was determined that the export log button was not in scope given some of the ONC security concerns that that introduces.

Q: I think my only direct design question concerns what should exactly appear, if anything, when someone clicks on a specific audit log, especially if we are not supporting export? Would we want some kind of modal giving more info about the action partaken? Bob also raises the idea of having an accordion dropdown. Either way, direction on how to tackle the next ticket would be appreciated!

A: Work to be addressed with updated design: https://linear.app/skylight-cdc/issue/QUE-203/frontend-audit-log-modal-or-dropdown

Relating to design, just noting that there are no tests for date; this is because date range will be changed in favor of date range picker to reflect the fact that users will want logs from a particular range of days, rather than just one date every time: https://linear.app/skylight-cdc/issue/QUE-208/frontend-implement-new-daterangepicker-design-for-audit-log.

## Related Issue

Fixes #353 

## Additional Information

## Checklist

- [x] Link to relevant issues
- [x] Update documentation


query-connector/src/app/(pages)/auditLogs/page.tsx
```diff
@@ -0,0 +1,255 @@
...
+            <tr>
+              <th className={styles.tableHeader}>Name</th>
+              <th className={styles.tableHeader}>Action</th>
+              <th className={styles.tableHeader}>Date</th>
+            </tr>
+          </thead>
+          <tbody>
+            {paginatedLogs.map((log, index) => (
+              <tr key={index}>
+                <td>{log.name}</td>
```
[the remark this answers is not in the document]
fzhao99: This was my bad for not tracking the designs closely enough, but I think we'll need to add a `message` column that gives folks information about the underlying function params / responses for audited functions. Current plan is to store it as JSON, which I'm guessing we can display raw in the table and maybe beautified in whatever modal we add on click?

cc: @mikang

query-connector/src/app/shared/page-routes.ts
```diff
@@ -51,6 +52,13 @@ pagesConfig[PAGES.FHIR_SERVERS] = {
   roleAccess: [UserRole.SUPER_ADMIN],
 };
 
+pagesConfig[PAGES.AUDIT_LOGS] = {
+  position: 3,
+  path: PAGES.AUDIT_LOGS,
+  name: "Audit Logs",
```
fzhao99, line 58: think we want lower case L in "Logs" to match the other sentence cased links on the page ("My queries", "FHIR servers")

query-connector/src/app/(pages)/auditLogs/page.tsx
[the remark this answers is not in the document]
fzhao99: an example from Aidbox for reference

https://github.com/user-attachments/assets/d0e48868-50f7-4b0e-9e24-39f498f93fb9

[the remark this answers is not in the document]
robertandremitchell: makes sense to me re: the fixed height. I assume we would just want to no-wrap the date field, right? I had originally no-wrapped all cells but because name/action could contain more dynamic lengths since they include patient data, query names, etc., I'm assuming we don't want to wrap those? We could also ellipsis long actions, I suppose

Screenshot 2025-03-13 at 9 58 48 AM

query-connector/src/app/shared/page-routes.ts
robertandremitchell: updated!

query-connector/src/app/(pages)/auditLogs/page.tsx
```diff
@@ -0,0 +1,255 @@
...
+          value={search}
+          onChange={(e: ChangeEvent<HTMLInputElement>) =>
+            setSearch(e.target.value)
+          }
+          className={styles.searchField}
+        />
+      </div>
+
+      <div className={styles.auditTableContainer}>
+        <Table className={styles.auditTable}>
```
fzhao99, line 196: This was my bad for not tracking the designs closely enough, but I think we'll need to add a `details` column that gives folks information about the underlying function params / responses for audited functions (ie FHIR query params). Current plan is to store it as JSON, which I'm guessing we can display raw in the table and maybe beautified in whatever modal we add on click?

an example from Aidbox for reference

https://github.com/user-attachments/assets/d0e48868-50f7-4b0e-9e24-39f498f93fb9

cc: @mikang

query-connector/src/app/(pages)/auditLogs/page.tsx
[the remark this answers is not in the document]
fzhao99: yeah I think ellipses makes sense past a max width (though the date entries will be Postgres NOW() values, so theoretically should be consistent). 

Per the other comment, think we'll need a details modal view, so in the event that things are too long for display we can display the full string there

query-connector/src/app/(pages)/auditLogs/page.tsx
fzhao99: totally fine to address this in a followup btw since it didn't make its way into the first ticket

query-connector/src/app/(pages)/auditLogs/auditLogs.module.scss
```diff
@@ -0,0 +1,126 @@
...
+  align-items: center;
+  gap: 0.5rem;
+  margin-top: 1.75rem;
+  margin-left: 4.25rem;
+}
+
+.tableHeader {
+  flex: 1 0 0;
+  color: var(--text-secondary, var(--gray-500, #919191));
+  font-size: 0.75rem;
```
johanna-skylight, line 57: The text for the table headers displays oddly small I think this style is causing it.

johanna-skylight: Why do we need customization to the header of the table in this page in general?

query-connector/src/app/(pages)/auditLogs/auditLogs.module.scss
```diff
@@ -0,0 +1,126 @@
...
+
+.auditTableContainer {
+  justify-content: space-between; 
+  align-items: center;
+  width: 70rem;
+
+}
+
+.auditTable {
+display: flex;
```
johanna-skylight, line 19: Why is this table style needed? in theory the table component now should be able to render any content dynamically without custom style.

johanna-skylight (changes requested):
Looks good overall. I left a comment regarding the table styles. I think with the latest refactor to the `Table` component no special style is needed per table per page.

query-connector/src/app/(pages)/auditLogs/auditLogs.module.scss
robertandremitchell: The header updates were just to match the figma design: https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=2476-2466&m=dev.

query-connector/src/app/(pages)/auditLogs/auditLogs.module.scss
robertandremitchell: you are correct, it shouldn't be. removing!

query-connector/src/app/(pages)/auditLogs/page.tsx
```diff
@@ -0,0 +1,259 @@
...
+    return filteredLogs.slice(
+      (currentPage - 1) * actionsPerPage,
+      currentPage * actionsPerPage,
+    );
+  }, [filteredLogs, currentPage, actionsPerPage]);
+
+  const totalPages = Math.ceil(filteredLogs.length / actionsPerPage);
+
+  return (
+    <div className={classNames(styles.mainContainerWider)}>
```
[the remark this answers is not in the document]
robertandremitchell: done!

mikang:
Thanks, everyone!

As discussed on our live Zoom call @robertandremitchell , here are the changes we landed on:

- Table headings as 14px
- Font color in selected item in pagination component as white
- When clicking on the date input field, open the date picker
- Background for “Search name or action” to be white
- Anchor the pagination component to the bottom - always 24px above the footer. If the table is long, then 24px of space between the table and the pagination component
image


- If no results found, then “No actions found” as text. Button only if there are filters applied (i.e. if there are no actions at all, then there is no “Clear filters” button) 
image

mikang:
whoops, that last one should be this 😄 

image

mikang:
@robertandremitchell we can follow the patterns here for the date picker. This is the official Salesforce design system. (The OG Salesforce app is not appealing for sure, but somehow, they have a nice design system — perhaps for their newer web apps?) You can try interacting with this date picker.

https://www.lightningdesignsystem.com/2e1ef8501/p/531362-datepicker

You can still type anything, but if you then press enter, then it gives an error message. 

image

query-connector/src/app/(pages)/auditLogs/page.tsx
robertandremitchell: created a follow-up ticket to pick up once we settle on an updated design: https://linear.app/skylight-cdc/issue/QUE-203/frontend-audit-log-modal-or-dropdown

johanna-skylight (approved):
The page and the code looks super clean! 🥇  awesome work Rob!!!

query-connector/src/app/(pages)/auditLogs/AuditLogs.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+ * @param text - The text to input
+ */
+const typeInField = async (placeholder: string, text: string) => {
+  const input = screen.getByPlaceholderText(placeholder);
+  await userEvent.type(input, text);
+};
+
+describe("AuditLogs Component", () => {
+  beforeEach(() => {
+    render(<AuditLogs />);
```
fzhao99, line 38: suggestion: if we could refactor this slightly to use the `user.[testEvent]` pattern [set up in other tests](https://github.com/CDCgov/dibbs-query-connector/blob/04b39beeb373f53e12670b587ac71876f57b5383/query-connector/src/app/(pages)/queryBuilding/buildFromTemplates/BuildFromTemplates.test.tsx#L54) to replace the instances of `fireEvent` / `userEvent` throughout the tests, this would make the tests more [in line with best practice](https://testing-library.com/docs/user-event/intro/). As is, using `fireEvent` may miss some interaction corner cases

query-connector/src/app/(pages)/auditLogs/page.tsx
```diff
@@ -0,0 +1,328 @@
...
+    () =>
+      logs.length > 0
+        ? new Date(Math.max(...logs.map((log) => log.date.getTime())))
+        : null,
+    [logs],
+  );
+
+  useEffect(() => {
+    const dateInput = document.querySelector(
+      ".usa-date-picker__external-input",
```
fzhao99, line 92: we may want to use a querySelector here that isn't a class / is more stable to CSS changes

fzhao99 (approved):
left two comments, but otherwise lgtm! if you get the design ok, feel free to merge and address in a followup since I know this has been out for a while

query-connector/src/app/(pages)/auditLogs/AuditLogs.test.tsx
robertandremitchell: makes sense. updated!

nickclyde (approved):
Awesome work Rob!!

PR #442, opened 2025-03-19
add extra note for local e2e

fzhao99:
Add an extra note to our readme for the aidbox license needed to get the seeder to run


robertandremitchell (approved):
good catch!

PR #453, opened 2025-03-24
jd/implement handling expired tokens

johanna-skylight:
# PULL REQUEST

## Summary

This work handles token expiration. For the server side the session is returned as null whenever the token is expired. In the client side (frontend) the time to expire is read upon first loading and a timeout is set to log out the user automatically.

## Related Issue

Fixes [QUE-204](https://linear.app/skylight-cdc/issue/QUE-204/implement-handling-of-token-expiration)

## Additional Information

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation


query-connector/src/app/shared/DataProvider.tsx
```diff
@@ -63,6 +63,7 @@ export function DataProvider({
       <SessionProvider
         session={session}
         refetchInterval={REFRESH_INTERVAL_MINS * 60}
+        refetchOnWindowFocus={false}
```
johanna-skylight, line 66: The app was triggering a call to the api/auth endpoint to resync the session object but the 15 mins interval should address any sync. Disabling resync on focus to save some HTTP requests.

query-connector/src/@types/next-auth.d.ts
```diff
@@ -1,15 +1,29 @@
...
+  interface JWT {
+    id?: string;
+    username?: string;
+    firstName?: string;
+    lastName?: string;
+    role?: string;
+    expiresIn?: number;
+  }
+}
+
```
johanna-skylight, line 28: I'm extending the token interface with the user fields

query-connector/src/app/shared/DataProvider.tsx
fzhao99: I know this isn't the first place where we added `REFRESH_INTERVAL_MINS`, but could we add a comment explaining how we picked 15 minutes (even if that comment is "it's arbitrary, Johanna just picked 15"? Could see a future where this figure is questioned for our compliance with some security protocol, and ideally would like to document the thought process to make sure future us has whatever context we can give them

johanna-skylight: done!

robertandremitchell (approved):
lgtm!

PR #481, opened 2025-04-02
refactor query execution

fzhao99:
# PULL REQUEST

## Summary
Refactoring query execution to support the audit log development work.

## Related Issue
In the work for [QUE-191](https://linear.app/skylight-cdc/issue/QUE-191/annotate-external-method-calls-with-auditable-decorator), it was discovered that the old setup for the query execution flow was passing around PHI in the params of the methods being called for query execution. Since our audit decorators store the contents of the function params for decorated functions, this resulted in us storing PHI. 

This PR refactors the query execution flow (which was something that we probably needed to do for a while now) to make the query functions stateless rather than building the query successively via arguments.

- Break apart the old `makeFhirQuery` function into patient discovery and (what we decided to call) patient records queries
- Move the query service methods into the service class pattern
- Change the typing / return values to accommodate the above

## Additional Information
Should be a no-op in terms of functionality, but please QA the query execution flow to make sure we haven't broken anything! 

## Checklist
- [X] Descriptive Pull Request title
- [X] Link to relevant issues
- [X] Ensure test coverage is above agreed upon threshold


robertandremitchell (commented):
lgtm!

however, I'm getting 
```
 POST /query 500 in 52ms
 ⨯ src/app/shared/fhirClient.ts (21:13) @ new FHIRClient
 ⨯ Error: No configuration found for server: Aidbox
    at new FHIRClient (./src/app/shared/fhirClient.ts:21:19)
    at QueryService.prepareFhirClient (./src/app/shared/query-service.ts:308:16)
    at async patientDiscoveryQuery (./src/app/shared/query-service.ts:247:28)
digest: "1185603628"
  19 |
  20 |     if (!config) {
> 21 |       throw new Error(`No configuration found for server: ${server}`);
     |             ^
  22 |     }
  23 |
  24 |     this.serverConfig = config;
```

from a clean build, Aidbox doesn't seem to be populating? I don't understand why and was trying to trace back more yesterday, but still pretty befuddled...

robertandremitchell (approved):
lgtm!

I'm gonna keep playing around with `aidboxone:stable` and `JAVA_TOOL_OPTIONS: "-XX:UseSVE=0"`. Even on stable, the Java tool options is still giving me grief, but at least on stable, aidboxone seems to more consistently start. Opening separate PR to look at further...

PR #493, opened 2025-04-07
Adding audit log hash update

### LINKED ISSUE #476: Replace hash filler function with an actual hashing algo
## Description:

In our auditing work, [this backend function](https://github.com/CDCgov/dibbs-query-connector/pull/447/files#diff-f94527bb5aa9106aab0cd2e9c140f5e951a6aee50928f15686d3c2d88dadf187R103-R105) needs to be replaced with actual hashing code that generates a checksum based off the author, audit contents, and timestamp (which isn't currently a param in the function since you could probably get access to `NOW()` within the function body.

The ONC regs require us to use a SHA-2 family algo to do the hashing, so we should probably find a library that lets us do that.

## Acceptance Criteria:

Function for checksum generates a hash string rather than just returning the constant

robertandremitchell:
# PULL REQUEST

## Summary

This adds a hashing function to the audit log table that conducts an update of the audit log role.

The other option to do this would be to INSERT the datetime, as opposed to relying on postgres' NOW() function. We need the datetime for the checksum, and I was concerned that the slight differences in using a datetime computed by the function and one by postgres could lead to small deltas in time. It might be okay if there's a millisecond difference, though...not sure.

## Related Issue

Fixes #476 

## Additional Information

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation

### DIFF
--- a/query-connector/src/app/auditLogs/decorator.ts
+++ b/query-connector/src/app/auditLogs/decorator.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { getDbClient } from "../backend/dbClient";
-import { generateAuditValues, generateAuditChecksum } from "./lib";
+import { generateAuditValues } from "./lib";
 
 export const AUDIT_LOG_MAX_RETRIES = 3;
 /**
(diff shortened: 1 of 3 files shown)


query-connector/src/app/auditLogs/decorator.ts
```diff
@@ -25,20 +25,42 @@ export function auditable(
...
+            author,
+            methodName,
+            JSON.stringify(auditMessage),
+          ]);
+
+          const insertedRow = result.rows[0];
+          const timestamp = insertedRow.created_at.toISOString();
+
+          // Step 2: Compute checksum with real timestamp
+          const checksum = generateAuditChecksum(
```
fzhao99, line 53: I think we should let JS generate the datestring. Think technically doing it this way is closer to the time of the audit event (ie the viewing of the record), plus it saves us a trip to the db.

Not super concerned about either one, but since we're breaking a tie the JS option is slightly preferable

nickclyde: Yeah, I agree - then I think we should be able to avoid multiple DB queries as well. EG, just generate a timestamp, then the checksum, then insert everything in one query.

robertandremitchell: makes sense. updating. just having some issue getting tests to run/pass. but 🤞🏽

nickclyde (approved):
:shipit:

PR #512, opened 2025-04-09
Fix "disable_cert_validation"

nickclyde:
# PULL REQUEST

## Summary

Fixes an issue with our `disable_cert_validation` config option for FHIR servers which was introduced when we upgraded to Node 22 - the new fetch implementation doesn't support the custom `https.Agent`. Instead, we now have to conditionally set the `NODE_TLS_REJECT_UNAUTHORIZED` env var to `0` based on the config option, and then reset it after the request is made.

### SSL Verification and Fetch Logic Refactor:

* [`query-connector/src/app/shared/fhirClient.ts`](diffhunk://#diff-a2a7ee99fa57e0b18a2959825565d2b1bc33fc1139df58a3cdb521924aee222cR6): Added a new import for `fetchWithoutSSL` from `./utils` to handle SSL verification.
* [`query-connector/src/app/shared/fhirClient.ts`](diffhunk://#diff-a2a7ee99fa57e0b18a2959825565d2b1bc33fc1139df58a3cdb521924aee222cR16): Updated the `FHIRClient` class to include a `fetch` function that conditionally uses `fetchWithoutSSL` based on the server configuration. 
* [`query-connector/src/app/shared/fhirClient.ts`](diffhunk://#diff-a2a7ee99fa57e0b18a2959825565d2b1bc33fc1139df58a3cdb521924aee222cL247-R244): Replaced all instances of `fetch` with `this.fetch` to ensure the correct fetch function is used throughout the class.

### Minor Fixes:

* [`query-connector/src/app/shared/query-service.ts`](diffhunk://#diff-e46929b05a2b4b65f4d05e65f1cf09e36d1c6c59118ecb577318c468e241a53fL367-R367): Fixed an issue in the `QueryService` class to correctly log the response body by awaiting `fhirResponse.text()`.


nickclyde:
@robertandremitchell Figured out the issue - we weren't passing the stored headers when doing the connection test. Most of the EHX servers should  be working now! Though CernerHelios seems to be throwing a 400, as it doesn't seem to support getting a Patient count 🤔

robertandremitchell (approved):
lgtm! thanks for chasing down the bug

PR #513, opened 2025-04-10
Initializing Audit Log Drawer

### LINKED ISSUE #434: Frontend: Audit Log Modal or dropdown
## Description:

We will need to add functionality to the row of each audit log to further drill down into the audit log for more details about the action taken.

Because we do not have the ability to export logs without greatly enhancing our initial burden for ONC compliance, it was determined that instead when someone clicks on the audit log, they should be able to review in greater detail the metadata that is associated with a particular action, since the display `action` should remain fairly general in the display.

We will need to determine:

* which actions to pull out of the `action` json (showing name and date would seem to make sense). This could potentially vary between what action is actually undertaken, since we might want to pull different metadata.
* as an initial first pass, once we determine what shape we would want (modal, accordion dropdown, other), it may make more sense to display the raw data formatted or copyable in some way for a user to look at more in-depth.
* This will likely require determining what action data we want, what action data users want.

## Acceptance Criteria:

robertandremitchell:
# PULL REQUEST

## Summary

Adds a drawer on click of individual audit row.

It also starts to build out infrastructure for rendering names instead of usernames for the name and also format our functions into readable, dynamic strings for the action column.

Screenshot 2025-04-11 at 11 17 15 AM
Screenshot 2025-04-11 at 11 17 22 AM

## Related Issue

Fixes #434 

## Additional Information

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation

### DIFF
--- a/query-connector/src/app/(pages)/auditLogs/page.tsx
+++ b/query-connector/src/app/(pages)/auditLogs/page.tsx
@@ -43,6 +43,7 @@ const AuditLogs: React.FC = () => {
     async function fetchAuditLogs() {
       const logs = await getAuditLogs();
       return logs.map((log) => {
+        // Convert createdAt to Date object that works in all browsers in correct timezone
         let str = String(log.createdAt).trim();
 
         if (!str.includes("T")) str = str.replace(" ", "T");


mikang:
Thanks, @robertandremitchell !! Here are some comments:
- Could you please remove the pagination component at the bottom if there are no results or only 1 page? It should appear only when there are 2 pages or more. 
image
- When hovering over a row in the audit log, 1) there should be a blue highlight and 2) the cursor should turn into a pointing hand to indicate that the row is clickable
- There should be no space between the "Full JSON request" and QC Admin, etc.

Regarding the date dropdown:
- "Clear" should be at the bottom of the dropdown left aligned
- There should be 16px of space between the two date pickers 

Currently it looks like this
image

robertandremitchell:
Screenshot 2025-04-14 at 1 44 32 PM

nickclyde (commented):
Amazing work, Rob! Ran it locally, really looks great. Mostly working as expected, though ran into two small bugs:

1. If I choose 50 per page, it works fine, but if I first go to the 2nd or 3rd page and then choose 50 per page, I end up with a blank page, all the table entries disappear
2. It's showing "Invalid date" for all of my entries - not sure if this is a me issue though? Maybe I need to clear my table?

robertandremitchell:
> Amazing work, Rob! Ran it locally, really looks great. Mostly working as expected, though ran into two small bugs:
> 
> 1. If I choose 50 per page, it works fine, but if I first go to the 2nd or 3rd page and then choose 50 per page, I end up with a blank page, all the table entries disappear
> 2. It's showing "Invalid date" for all of my entries - not sure if this is a me issue though? Maybe I need to clear my table?

ack, let me look at #1. I'm guessing it's being overzealous in the new code I added to hide pagination/actions per page.

re: #2, can you share a screenshot? The start/end dates are rather permissive on formats (MM/DD/YYYY, YYYY-MM-DD, etc.) so curious to see what's happening there.

nickclyde:
Screenshot 2025-04-15 at 1 10 22 PM

Sure thing, here's the screenshot, and here's what my db looks like:

Screenshot 2025-04-15 at 1 11 05 PM

robertandremitchell:
> Screenshot 2025-04-15 at 1 10 22 PM
> 
> Sure thing, here's the screenshot, and here's what my db looks like:
> 
> Screenshot 2025-04-15 at 1 11 05 PM

interesting...having trouble replicating locally but can keep looking. My initial thought is that it has to do with the code I added here:

```typescript
  useEffect(() => {
    async function fetchAuditLogs() {
      const logs = await getAuditLogs();
      return logs.map((log) => ({
        ...log,
        createdAt: new Date(log.createdAt + "Z"),
      }));
    }
```

@nickclyde , would you be able to test the code with this?

```typescript
  useEffect(() => {
    async function fetchAuditLogs() {
      const logs = await getAuditLogs();
      return logs;
    }
```

The reason I added the `Z` is that I found that when formatting the date below that it defaults to UTC offset 0 (GMT) because it interprets the Z-less date that way, while the Z-ful date will look at local system time and display the datetime in the appropriate timezone.

However, I wonder if this is not an OS-agnostic solution to this issue. If the original code works, I can try to find an alternative method for displaying datetimes in the user's timezone.

robertandremitchell:
@nickclyde , I pushed a more comprehensive rewrite of how to try to handle datetime stamps. lmk if you still run into `Invalid Date`; I'm hoping this should do the trick.

query-connector/src/app/(pages)/auditLogs/components/auditLogMaps.ts
```diff
@@ -0,0 +1,142 @@
...
+  fallback?: string,
+): string {
+  const full = `${first ?? ""} ${last ?? ""}`.trim();
+  return full !== "" ? full : (fallback ?? "");
+}
+
+/**
+ * Maps action types to their labels and formats.
+ */
+export const auditLogActionTypeMap: Record<string, auditLogActionTypeMapping> =
```
katyasoup, line 60: nice!

katyasoup (approved):
looks great! i'm not seeing any issues with the date formatting on my end

query-connector/src/app/(pages)/auditLogs/components/auditLogMaps.ts
fzhao99: +1!

Consideration for future improvement, no action needed: I'm a little nervous as we build out more of this feature that there's so much custom logic for transforming data values at the UI level, and depending on churn of this code, worried about DRYness. Think it might be overoptimizing to try and generalize things from here, but if we find ourselves updating this constant too much, would love to look into generalizing the formatting logic here (maybe by putting some stricter regs around the method names so that they're more consistently verbDescriptorObject?)

query-connector/src/app/(pages)/auditLogs/page.tsx
```diff
@@ -29,13 +36,26 @@ const AuditLogs: React.FC = () => {
...
+  const [selectedLog, setSelectedLog] = useState<LogEntry | null>(null);
 
   useEffect(() => {
     async function fetchAuditLogs() {
       const logs = await getAuditLogs();
-      return logs;
+      return logs.map((log) => {
+        let str = String(log.createdAt).trim();
+
+        if (!str.includes("T")) str = str.replace(" ", "T");
```
fzhao99, line 48: non-blocking suggestion: Could we add a comment here for why this is necessary? Guessing it's a timezone / stamp conversion thing, but want to be kind to future devs :)

query-connector/src/app/(pages)/auditLogs/page.tsx
```diff
@@ -47,11 +67,23 @@ const AuditLogs: React.FC = () => {
   const [filteredLogs, setFilteredLogs] = useState(logs);
 
   const uniqueNames = useMemo(
-    () => Array.from(new Set(logs.map((log) => log.author))).sort(),
+    () =>
```
fzhao99, line 70: note for future, no action needed: guessing these may become opportunities for performance tuning / better SQL as the audit table gets big, since the set operation in JS is pretty expensive. We should keep an eye on this!

query-connector/src/app/(pages)/auditLogs/page.tsx
robertandremitchell: makes sense, added! and yeah, it's for timezone display

fzhao99 (approved):
lgtm! great work on this rob :)

query-connector/src/app/(pages)/auditLogs/page.tsx
robertandremitchell: added backlog ticket: https://linear.app/skylight-cdc/issue/QUE-245/backlog-investigation-write-less-expensive-way-to-do-sortset-for-audit

query-connector/src/app/(pages)/auditLogs/components/auditLogMaps.ts
robertandremitchell: ticket here: https://linear.app/skylight-cdc/issue/QUE-246/spike-improvement-ideas-for-audit-log-mapping

nickclyde (approved):
Great work Rob, thanks for fixing the date issue!

PR #525, opened 2025-04-17
modularize e2e env var seeding

fzhao99:
# PULL REQUEST

## Summary
When looking into fixing the e2e's noticed some areas of improvement for how we're seeding env variables in our e2e scripts. 

- All Dockerized versions of the e2e's (in CI and local run using `npm run test:playwright`) have the e2e's seeded within the `.e2e_tests.sh` script rather than relying on the `ci.yaml` file and the dev's local .env file respectively
    - We do have one secret (the Aidbox license) in the e2e run chain, so added some code to look for that value in the `.env` and then seed that accordingly 
- The Playwright UI command has its e2e var's seeded in the `.e2e_tests_local.sh` script
- All e2e tests now generate / teardown their own `.env.e2e` files so as to separate concerns with any floating .env files
- Added a note about the SMART on FHIR e2e not working for local UI mode because of Docker networking issues

## Checklist

- [x] Update documentation

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -91,24 +91,24 @@ jobs:
       - name: Create .env
         working-directory: ./query-connector
         run: |
-          echo "AIDBOX_LICENSE=${{ secrets.AIDBOX_LICENSE }}" > .env.e2e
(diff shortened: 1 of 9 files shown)


robertandremitchell (approved):
looks good to me! 

not blocking; not sure what happened but I got smart on fhir to pass the second time I ran it? 🤷🏽

PR #530, opened 2025-04-21
Add version numbers to query connector packages

shanice-skylight:
# PULL REQUEST

## Summary

- Added code to add version number[ shorthand for {{major}}.{{minor}}.{{patch}}] to each package. This is trigger on a merge to `main` 
- The version number will allow users to specify what version of the application they would like to run and is a dependency for dibbs-vm
- Package version starts at 0.9.0 based on this [thread](https://skylight-hq.slack.com/archives/C05BEG226RZ/p1744653382463439)

- **Details on worklow:** 

  1. Get Current QC Version: The current version is retrieved from the app_version.yml file.
  2. Increment QC Version: A patch version bump is applied to the current version, unless a different version is provided, in which case the new version is assigned to the new_version variable.
  3. Update app_version.yml: The new_version is written into the app_version.yml file.
  4. Extract Docker Metadata (Tags & Labels): Using docker/metadata-action, the new_version is extracted and tagged as both the new version and latest.
  5. Push Docker Images to Registry: The docker/build-push-action grabs the created tags and pushes the corresponding images to the container registry.


fzhao99 (approved):

robertandremitchell (approved):
makes sense to me. I guess my only question is it looks to auto-increment, so I think so long as we review PRs that's fine.

shanice-skylight:
> makes sense to me. I guess my only question is it looks to auto-increment, so I think so long as we review PRs that's fine.

It automatically increments the patch version based on the version in the` app_version.yml `file. You are correct we do have to make sure we review this. Alternatively, I could disable the auto-increment, and we would need to update the version manually during PR changes whichever works best for the teams workflow.

PR #562, opened 2025-04-25
Change aidbox client name to query-connector

nickclyde:
# PULL REQUEST

## Summary
The seed script is inserting the Aidbox `fhir_servers` table entry with a `client_id` of "query-connector", but the script is  setting up the client in Aidbox with the id "e2e-smart-test-client", so a fresh spin up will result in failures when trying to talk to Aidbox. This should resolve that by making them consistent!


fzhao99 (approved):
Nice catch!

robertandremitchell (approved):

Issue #546, opened 2025-04-22
Add "manage codes" page - edit codes/"select" state


katyasoup:
## Description:

Add a page for managing value set library with a button for adding new value sets

[https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=3166-3358&t=kFnELQmjBZHzcy6T-1](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=3166-3358&t=kFnELQmjBZHzcy6T-1)

## Acceptance Criteria:

* Builds off of [QUE-230](https://linear.app/skylight-cdc/issue/QUE-230/add-manage-codes-page-read-onlymanage-state)
* Adds "select" state where user can add codes to a query, rather than read-only state
* It also allows users to click a button to edit value sets that they created. Valuesets from CSTE will be locked

Issue #503, opened 2025-04-08
String all the new pages for custom codes together


nickclyde:
## Description:

Implement the complex routing needed for custom codes

## Acceptance Criteria:

* hen a user clicks the button in query building to add additional codes, go to the "select codes" page
* From there, a user can select existing valuesets and add them to the query, returning to the query building page
* They can also click "manage codes" to go to the "manage codes" page, which will trigger a save of the query and a toast
* From "manage codes", the user should be able to go back to the query building page
* They should also be able to click a button to add a value set
* After adding a value set, it should go back to "manage codes"

robertandremitchell:
To support seamless tracking of the currently edited query across `/queryBuilding` and `/codeLibrary`, we propose storing the `queryId` both in the URL (as a `?queryId=` param) and in the shared `DataContext` (`selectedQuery.queryId`). This dual tracking ensures that state is preserved on navigation, supports deep-linking and bookmarking, and enables the code library to append custom codes to the correct query. We’ll update the `handleSaveQuery` logic to populate `selectedQuery`, add a reusable redirect helper to push to `/codeLibrary?queryId=...`, and ensure `/codeLibrary` initializes context from the query string. This approach avoids restructuring existing logic while enabling accurate query association and a smoother user experience.<br><br>@katyasoup, curious your thoughts as you've been more involved with the frontend aspects of this. I could split the ticket to start on adding / confirming this functionality works with the /queryBuilding workflow before linking to /codeLibrary.

Issue #502, opened 2025-04-08
Add new value set page


nickclyde:
## Description:

Add a page for creating new value set

[https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=3166-3358&t=kFnELQmjBZHzcy6T-1](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=3166-3358&t=kFnELQmjBZHzcy6T-1)

## Acceptance Criteria:

* There is a page where a user can add a new value set

PR #573, opened 2025-04-29
Add Code Library page

### LINKED ISSUE #501: Add "manage codes" page - read only/manage state
## Description:

Add a page for managing value set library with a button for adding new value sets

[https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=3148-5175&t=ZUvf4iQvHSuxGm3e-0](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=3148-5175&t=ZUvf4iQvHSuxGm3e-0)

## Acceptance Criteria:

* A new static page built out for the "Manage codes" view linked above
* It should pull existing value sets from the db into a paginated table
* Search and drop-down filters update the view accordingly\*
  * \*Created by me / Created by my team maybe need to be handled in future ticket?
* Clicking a value set should render view-only details
  * Non-custom value sets should show a lock indicating they can't be edited
  * Edit capability to be handled in [QUE-257](https://linear.app/skylight-cdc/issue/QUE-257/add-manage-codes-page-edit-codes)
* "Add value set" button is present but functionality to be handled in [QUE-231](https://linear.app/skylight-cdc/issue/QUE-231/add-new-value-set-page)

katyasoup:
# PULL REQUEST

## Summary

Adds a new page `Code library` that is accessible from the header dropdown menu. It:
- loads all available value sets into a paginated table
  - renders a skeleton loading view while value sets are fetched
- fetches code/concept details that display to the right when a value set is clicked
- implements highlighter text search on value set name, condition name, category, and code system; implements filtering via a drop down menu on category, code system, and value set author

Not implemented/future work:
- Backnav link should be dynamic based on how you got to the page (#503 )
- "Add value set" button opens the custom value set page (#502 )
- "Select" mode enables adding value sets from the table to a query (#546 )
- Dropdown filter shortcuts "Created by me" and "Created by my team" to come once backend work is merged

## Related Issue

Fixes #501 #559 

## Additional Information
**Loading:**
codeLibrary-loading

**Loaded:**
codeLibrary-loaded

**Text search:**
codeLibrary-text-highlight-search

**Filters:**
codeLibrary-text_and_filters

**Pagination:**
codeLibrary-pagination

## Checklist

- [x] Link to relevant issues
- [x] Update documentation


src/app/shared/database-service.ts
```diff
@@ -563,6 +565,36 @@ class DatabaseService {
     }
   }
 
+  /**
+   * Retrieves all available value sets in Query Connector.
+   * @returns A list of value sets registered in the query connector.
+   */
+  static async getAllValueSets(): Promise<QCResponse<DibbsValueSet>> {
```
fzhao99, line 572: think we can use the `@adminRequired` decorator here since this is in the service class

fzhao99:
really impressive first pass Katie!

I'm noticing some strange rendering behavior in the case of the filters triggering an empty state.

1. If you have a valueset selected / with codes rendered and then apply additional filters on top of that, the selected conditions persist even though we've left the selected valueset state. Wondering if we want to reset the state of the concept codes when there's an update to the filter / the valuesets get regenerated
https://github.com/user-attachments/assets/9617dae3-1f96-41ca-bc94-e2f65d172825

2. In those cases, it seems the pagination minimum is 5. Guessing we want to lock that value to 1 when there aren't enough entries to paginate against 
https://github.com/user-attachments/assets/fc836593-f0df-4d78-b7ef-e0751c1c2836

robertandremitchell (approved):
this looks really awesome!

katyasoup:
>     1. If you have a valueset selected / with codes rendered and then apply additional filters on top of that, the selected conditions persist even though we've left the selected valueset state. Wondering if we want to reset the state of the concept codes when there's an update to the filter / the valuesets get regenerated

Yes, good call out! Would love @mikang's input here - I had set it up originally to remove the right-side details if you typed in the search and the selected value set was no longer a match. It was a little jarring to have it disappear, but I agree it's strange to have it persist like it does currently.

>     2. In those cases, it seems the pagination minimum is 5. Guessing we want to lock that value to 1 when there aren't enough entries to paginate against

Good catch, will update!

src/app/(pages)/codeLibrary/page.tsx
```diff
@@ -0,0 +1,491 @@
...
+      <div className={classNames("main-container__wide", styles.mainContainer)}>
+        <div className={styles.header}>
+          {mode === "manage" ? (
+            <Backlink onClick={() => {}} label={"Back to Query library"} />
+          ) : (
+            <Backlink onClick={goBack} label={"Back to My queries"} />
+          )}
+          <h1 className={styles.header__title}>Manage codes</h1>
+          <div className={styles.header__subtitle}>
+            Click on the checkbox to delete the value set or code
```
katyasoup, line 260: @mikang this is the copy from Figma but wondering if we want something different for this view-only/manage mode?

mikang:
Thank you @katyasoup !! LOVE the skeleton view and overall structure. Also how smooth/fast it is for the pagination component to update when filtering! I can see a lot of attention to detail here. 

Here is my feedback:

Questions
- [ ] How are the value sets sorted? Could we sort alphabetically? Including when it's filtered
- [ ] Do all value sets end with SNOMED/ICD10CM/etc.? If so, we could drop the code description in the "[Disease] · [Condition/Lab/Medication]" secondary line. If the user creates a custom value set, then we could append their value set name with "([Value set name])"

Interaction
- [ ] Upon load, auto-select the first value set. It should be highlighted in blue with the codes appearing on the right side
- [ ] Move "Code library" underneath "Query library" in the dropdown menu on the upper right hand corner
- [ ] "Back to Query library" backlink should be dynamic. For example, when I clicked on FHIR servers, then went to the Code library page, then it still showed "Back to Query library" — if I navigate to this page from anywhere else that is not the Query library, then the backlink should not appear. (This may be what was mentioned in the "Future work" section. If so, please disregard this c

ommment.)
- [ ] If there is only one page of results, remove "1" from the middle of the pagination component.
- [ ] Add "(filtered)" after the "Showing 1 - 10 of 591 value sets"
- [ ] If I filter, then the value sets column should update with the first entry selected (highlighted with the respective codes displayed on the right).
- [ ] If I filter and there are no value sets, it should say "No value sets found" on the left (as it is now) and no codes on the right. Right now, it looks like this:

image


Visual
- [ ] "Back to Query library" should be in the standard link color
- [ ] "Value sets are an organizational structure..." should be 16px
- [ ] "VALUE SET" column header should be 14px
- [ ] Center the lock icon so that it sit vertically in the center with "This value set comes from ..."
- [ ] Drop the (s) after Conditions/Labs/Medications (this is a change from the original design)
- [ ] Add 8px of whitespace below the "Code/Name" header so it provides some distance from the text underneath
- [ ] Left align the filters menu with the filter icon, not the "Filters" text
- [ ] If a value set comes from DIBBs, it should say, "This value set comes from DIBBs and cannot be modified" (not CSTE)
- [ ] "Start from scratch" for building a new query should always be the default link color, even if it was clicked on before. (Right now, it's purple if I already clicked on it once.)
- [ ] Bump up the pagination component by 16px. There should be 16px between the bottom of the pagination component and the top of the footer. This may result in a change across all pages.

katyasoup:
> Do all value sets end with SNOMED/ICD10CM/etc.? If so, we could drop the code description in the "[Disease] · [Condition/Lab/Medication]" secondary line. If the user creates a custom value set, then we could append their value set name with "([Value set name])"

@mikang no, not all of them end with that - that is the name of the value set as it is stored in the db, looks like it is used to differentiate value sets that otherwise would have the same name (see below - same idea as "Tests by Nucleic Acid" vs "Tests by Culture" vs "Tests by Antigen" which otherwise are all called the same thing)
Screenshot 2025-05-05 at 10 42 33


I think it's valuable to have the source system displayed in the row; we could probably write a formatter to strip the `(SNOMED)` etc from the name, but fwiw we do display both in the Query Library view:

Screenshot 2025-05-05 at 10 49 15

katyasoup:
>"Back to Query library" backlink should be dynamic. 

Will be handled in future ticket!

> Back to Query library" should be in the standard link color
> Left align the filters menu with the filter icon, not the "Filters" text

These both show correctly on my end, can you add a screenshot showing the incorrect color/alignment?

Everything else should be updated in this PR now! @mikang

nickclyde (approved):
Fantastic work Katie!!

PR #574, opened 2025-04-30
Initializing custom code backend

### LINKED ISSUE #505: Add backend logic for adding valuesets
## Description:

Add a database function for adding valuesets

## Acceptance Criteria:

* There is a service function with upserts codes to the database, adds a new value set, and links the codes to the valueset
* Valuesets should be created with a UUID or maybe something else

robertandremitchell:
# PULL REQUEST

## Summary

The new class seeks to:
1. Initialize a custom code condition. This condition will essentially hold all the custom valuesets users generate, connecting them all to this to allow for a more seamless use of our existing condition->valueset-> concept valueset.
2. Add a async that adds valueset created by user and connects it to their concepts. 
3. Add integration test to confirm adding valueset and concepts works.

## Related Issue

Fixes #505 

## Additional Information

There's a few backend design considerations for the next step: custom code condition will need to have access to ALL valuesets. I.e., someone may have a Newborn Screening query, and they might wholesale add a new valueset or pull from an existing valueset. That existing valueset has to be able to appear on the custom code section of the page.

One idea I had was at the initialization of custom code condition that it just simply load all the existing valuesets into its crosswalk table; this would then allow a user to enable and disable which ones they want.

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2975,6 +2975,19 @@
         "node": ">= 10"
       }
     },
+    "node_modules/@noble/hashes": {
+      "version": "1.8.0",
+      "resolved": "https://registry.npmjs.org/@noble/hashes/-/hashes-1.8.0.tgz",
+      "integrity": "sha512-jCs9ldd7NwzpgXDIf6P3+NrHh9/sD6CQdxHyjQI+h/6rDNo88ypBxxz45UDuZHz9r3tNz7N/VInSVoVdtXEI4A==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": "^14.21.3 || >=16"
+      },
+      "funding": {
+        "url": "https://paulmillr.com/funding/"
+      }
+    },
     "node_modules/@nodelib/fs.scandir": {
       "version": "2.1.5",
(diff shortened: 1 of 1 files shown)


src/app/shared/custom-code-service.ts
```diff
@@ -0,0 +1,136 @@
...
+import {
+  INTENTIONAL_EMPTY_STRING_FOR_CONCEPT_VERSION,
+  INTENTIONAL_EMPTY_STRING_FOR_GEM_CODE,
+} from "./constants";
+import type { DibbsValueSet } from "../models/entities/valuesets";
+import crypto from "crypto";
+
+export class UserCreatedValuesetService {
+  private static get dbClient() {
+    return getDbClient();
```
fzhao99, line 22: could we import / use the dbService from `@/app/backend/dbServices/db-service` instead here? that service is just a wrapper around this imported client but has the camel case annotation on it.

Would love for us eventually to consolidate both db setup files into one (which I think is still undoable because of some leftover DB interactions in the db generation file)

robertandremitchell: updated!

katyasoup (approved):
🎉

fzhao99 (approved):

PR #577, opened 2025-05-05
mark user management functions as auditable

fzhao99:
# PULL REQUEST

## Summary
Marks user management / session management functions as auditable, as well as some other improvements (see additional info)

https://github.com/user-attachments/assets/b38ae2d0-b67a-4b94-9def-f623c3631bc3

## Related Issue
Fixes [QUE-194](https://linear.app/skylight-cdc/issue/QUE-194/methods-handing-user-session-management-are-marked-as-auditable)

## Additional info
- Extends the camel case decorator to handle nested objects and stringified DB stores, as well as arrays and circular references
- Adds maps for the session management functions. Will add some tests for these in an upcoming PR. Reorganize the typing a little bit to be more generic
- Refactors the header to take session information as a prop to hide / render buttons correctly when the user logs in or out
- Puts a wrapper around the session management (sign in / out) functions that allow things to be auditable
- Refactors the audit logic to expose a success callback method that the tests await on to make them less prone to race conditions

## Checklist

- [x] Link to relevant issues
- [ ] Update documentation


src/app/backend/usergroup-management.ts
```diff
@@ -241,59 +241,6 @@ class UserGroupManagementService {
...
 
-  /**
-   * Saves the user group membership for a specific user group.
-   * @param groupId - The unique identifier of the user group.
-   * @param selectedUsers - The unique identifiers of the users to add to the group.
-   * @returns A list of user group memberships for the specified group.
-   */
-  @transaction
-  @adminRequired
-  static async saveUserGroupMembership(
```
fzhao99, line 252: this function wasn't getting used anywhere (since I think we're just using the add / remove functions directly) so I went ahead and deleted it

robertandremitchell (commented):
looks good to me! structure overall makes sense, though I do wonder if we would want to capture the actual user's names and group names in the topline level of the audit log? Especially in the event of a user group deletion, it might be harder to trace back which group was deleted with just the ID.

fzhao99:
> I do wonder if we would want to capture the actual user's names and group names

this is a little more complicated to do since that info isn't captured in the function inputs. think you'd be able to recover that information by searching for the ID of the newly deleted group and then determine the chain of additions / updates that maps to the group information at the time of deletion. From what I understand this is what we'd need for compliance, but lmk if you think we should look into developing this out further to include that info

src/app/backend/dbServices/decorators.ts
```diff
@@ -142,11 +142,7 @@ export function camelCaseDbColumnNames<T extends Record<string, unknown>>(
       }
 
       result.rows = result.rows.map((v) => {
-        const val: Record<string, unknown> = {};
-        Object.entries(v).forEach(([k, v]) => {
-          val[translateSnakeStringToCamelCase(k)] = v;
-        });
-
+        const val = translateNestedObjectKeysIntoCamelCase(v as object);
```
nickclyde, line 145: Nice, much cleaner

src/app/tests/integration/audit-logging.test.ts
```diff
@@ -165,26 +154,91 @@ describe("audit log", () => {
...
     ).rejects.toThrow(/UPDATEs to audit_logs are not permitted/i);
-
     await expect(
       dbClient.query(`DELETE FROM audit_logs WHERE id = $1`, [insertedId]),
     ).rejects.toThrow(/DELETEs from audit_logs are not permitted/i);
   });
 });
 
+describe("sign in and out", () => {
+  it("sign out should generate an audit entry", async () => {
```
nickclyde, line 174: 👍🏻

nickclyde (approved):
Looks good, thanks Bob!!

PR #588, opened 2025-05-06
Add Value Set page

katyasoup:
# PULL REQUEST

## Summary
- Adds a new page view to Add or Edit a custom value set and codes.
- Connects to the backend to save and update value sets, concepts, and all associated joins
- Implements "Created by me" and "Created by my team" filters

**Not included/future work:**
- Delete functionality (#587)
- Form validation (#504)
- Adding custom value sets/codes to a query (#546)
- Additional details/cleanup to be handled in #503 

## Related Issue

Fixes #502 

## Additional Information
- Would love extra eyes on how we're loading/refreshing the entire list of value sets when toggling between views - it feels slow and redundant to re-fetch everything, but we also should be able to see our new/updated data when we switch back to the table view from the form.

## Checklist

- [x] Descriptive Pull Request title
- [x] Link to relevant issues
- [x] Provide necessary context for design reviewers
- [ ] Ensure test coverage is above agreed upon threshold
- [ ] Update documentation

PR #589, opened 2025-05-07
reorganize backend db services files

fzhao99:
# PULL REQUEST

## Summary
Some cleanup to make the file naming conventions consistent / to collapse some duplicate files into one. Part of the work for audit logs now that there's enough functionality to warrant a standard.

## Additional Information
- Files in `backend` are now kebab-cased to indicate they're server-side functionality. 
- For any files that need to be split into separate files (mainly for import fiddliness in testing files), folders are also kebab cased and sub files are named like `lib.ts`, `service.ts` or `decorator.ts` so that the imports are relatively similar to the ones that don't need folders
- Put all the database client / config code into one folder and added a note / indication via an `internal_` prefix to hopefully point future devs to the right database client to import
- Updated imports / references accordingly

## Checklist


robertandremitchell (approved):
makes sense to me. i don't have any big PRs I'm trying to land today but may be worth waiting to hear from @nickclyde and @katyasoup before a merge.

katyasoup (approved):
lgtm! if we can wait until #588 merges that would be ideal - hoping to get that ready for :eyes today

PR #590, opened 2025-05-07
mark query building actions as auditable

fzhao99:
# PULL REQUEST

## Summary
Adds auditability decorator and tests to the relevant query building methods

## Related Issue
Fixes [QUE-193](https://linear.app/skylight-cdc/issue/QUE-193/methods-for-query-building-marked-as-auditable)

## Checklist
- [x] Link to relevant issues


src/app/(pages)/auditLogs/components/auditLogMaps.ts
robertandremitchell, line 1: can we update this line:

`const username = request?.profile.preferredUsername;`


to

`const username = request?.profile?.preferredUsername;`

it crashes for me without that.

src/app/(pages)/auditLogs/components/auditLogMaps.ts
```diff
@@ -253,6 +253,23 @@ export const auditLogActionTypeMap: Record<string, auditLogActionTypeMapping> =
...
+        return `Query with ID ${queryId} deleted`;
+      },
+    },
+    saveCustomQuery: {
+      label: "Added a new query",
+      format: (log) => {
+        const request = parseRequest(log);
+        const queryName = Object.values(request)[1];
+        const author = Object.values(request)[2];
+        return `Query with name ${queryName} made by ${author}`;
```
robertandremitchell, line 270: I would maybe say "saved" instead of "made" here, since a log entry is made upon create or update. I don't think we necessary need to disambiguate since if there's a question about a change log, we will have saved all versions of the query.

src/app/(pages)/auditLogs/components/auditLogMaps.ts
```diff
@@ -253,6 +253,23 @@ export const auditLogActionTypeMap: Record<string, auditLogActionTypeMapping> =
...
+        const request = parseRequest(log);
+        const queryId = Object.values(request)[0];
+        return `Query with ID ${queryId} deleted`;
+      },
+    },
+    saveCustomQuery: {
+      label: "Added a new query",
+      format: (log) => {
+        const request = parseRequest(log);
+        const queryName = Object.values(request)[1];
```
robertandremitchell, line 268: I would save name and queryId. id could be helpful for searching/parsing through a log to get the history of a query imo.

robertandremitchell (commented):
lgtm! a few ticky tackys I think might be worth addressing

src/app/(pages)/auditLogs/components/auditLogMaps.ts
robertandremitchell: additionally, we shouldn't need author name in the action since it's gonna be saved under the name column imo

robertandremitchell (approved):

nickclyde (approved):

PR #601, opened 2025-05-12
Adding save query to condition selection step of Query Building and assign authorship

### LINKED ISSUE #555: Auto-save the query when it is created
## Description:

After adding a query name and selecting template checkboxes (or adding custom codes), automatically save the query when they get to the next page.

## Acceptance Criteria:

robertandremitchell:
# PULL REQUEST

## Summary

This does two things:

* Adds the saved query once condition(s) have been selected and a name has been selected
* Adds the user.id as the `author` instead of hardcoded DIBBs.

Open questions:
* Do we envision a scenario where we ever update author? Or is it hard-set from the initial creator even if someone else goes in and does a bunch of edits?
* Do we want to save user.id as author? I don't think we ever present this info anywhere, and we can obv connect to users if we need to get first/last name.
* There's a somewhat larger convo about what should happen when we hit `start from scratch` on this page. One idea I had was that we should still create a query.id at this step and save it at this step. We should be able to save this onto the codeLibrary page, but I'm wondering if we want to break convention / provide greater protections for ourselves by having `codeLibrary/queryId={}` as the standard. In the reading I've done, we shouldn't technically need this, but saving the ID with just session info does potentially leave us vulnerable to users bouncing between `queryBuilding` and `codeLibrary` by giving us the ID in the URL as a fallback param. Curious if that is something we would want to start adding here for the queryBuilding and codeLibrary workflow. Also happy to keep noodling on this, I totally didn't pick up this ticket as a backdoor way to start this conversation 😅 

## Related Issue

Fixes #555 

## Additional Information

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation


nickclyde (approved):
Looks great!

> Do we envision a scenario where we ever update author? Or is it hard-set from the initial creator even if someone else goes in and does a bunch of edits?

I think it's fine to keep the author as the OG for now until somebody asks for different behavior

> Do we want to save user.id as author? I don't think we ever present this info anywhere, and we can obv connect to users if we need to get first/last name.
Yeah, I think user.id makes the most sense.
> There's a somewhat larger convo about what should happen when we hit start from scratch on this page. One idea I had was that we should still create a query.id at this step and save it at this step. We should be able to save this onto the codeLibrary page, but I'm wondering if we want to break convention / provide greater protections for ourselves by having codeLibrary/queryId={} as the standard. In the reading I've done, we shouldn't technically need this, but saving the ID with just session info does potentially leave us vulnerable to users bouncing between queryBuilding and codeLibrary by giving us the ID in the URL as a fallback param. Curious if that is something we would want to start adding here for the queryBuilding and codeLibrary workflow. Also happy to keep noodling on this, I totally didn't pick up this ticket as a backdoor way to start this conversation 😅
Yeah, I think we should save as soon as we go to the value set selection screen, whether it uses conditions or starts from scratch. You're right that it will make navigating back to queryBuilding from the codeLibrary easier!

PR #604, opened 2025-05-13
Add ability to delete custom codes and value sets

### LINKED ISSUE #587: Delete custom codes and value sets
## Description:

Add the ability to delete custom codes from custom value sets, as well as delete whole custom value sets.

A few things to consider:

* If a custom value set exists, should we prevent the user from deleting *all* codes? I.e. if you delete the only code in the value set, either throw an error or ask if they want to delete the entire value set?
* How to handle if the code is used in more than one value set? Should deleting the code only remove the valueset-code association? Should the user be asked if they want to remove the code everywhere?
  * Inverse of the above - does deleting a custom value set delete the custom codes associated with it? **If not**, there is not currently a way to manage those codes that are not associated with a value set (unless a user knows the exact ID and name and they are added to a new custom value set)

## Acceptance Criteria:

* Answers to the above qs are clarified
  1. **No, allow custom value sets to be empty**
     1. empty state mockup [here](https://www.figma.com/design/Iuw9me6kYftBF4WTCEsCZz/Query-Connector?node-id=3654-2288&t=0hghUoebdqzUT24A-0) 
  2. **Deleting a value set does not delete the codes, just the association**
* Users can delete custom codes and value sets in accordance with above decision
  * Toasts and confirmation modals are added if applicable

katyasoup:
# PULL REQUEST

## Summary
Adds ability to delete custom codes and value sets:

Deleting a value set:
- A button should display in the right-side panel when a user clicks on a custom value set
- A modal will ask the user to confirm they want to delete the value set
- On confirmation, the modal closes, a toast is shown with a success/error message, and value sets are refreshed; any filters are cleared

Deleting a code:
- When adding a new value set, or when editing an existing value set, a user can hover over the row of inputs and click the trash icon that appears
- No modal/confirmation is required to delete a code. A success/error toast appears after the action.
- You should not need to click the "Save" button for the delete to persist. (You DO need to click Save to add/update a code)

Limitations/not in scope:
- There are currently no authorship restrictions for deleting codes/value sets (eg. limiting a user to only deleting value sets they themself created). There ARE admin checks required for all methods related to this epic (viewing the page, insert/update/fetch/delete methods, etc)

## Related Issue

Fixes #587 

## Additional Information
I think I got all the funky behavior sorted out when you add/edit/delete lots of codes to a a new/existing value set, but please play around and try to break it some more 😉 

## Checklist

- [x] Link to relevant issues
- [ ] Update documentation


robertandremitchell (approved):
looks good to me! Tried a few stress tests and seemed good.

PR #605, opened 2025-05-13
Tracking selectedQuery with Data Context

robertandremitchell:
# PULL REQUEST

## Summary

Promoting selectedQuery as an optional parameter in DataContext with the goal of being able to track its context across multiple pages. We will focus on updating `/queryBuilding` change to ensure it is present/updating properly before looking at how to ensure it pushes to `/codeLibrary`.

I added a dummy box at the top of /queryBuilding and /codeLibrary to show how it looks/works at each state.

**Just starting on /codeLibrary**
Screenshot 2025-05-14 at 12 25 16 PM

**Hitting the Start from Scratch Button**
https://github.com/user-attachments/assets/3f83158f-5937-4b1c-9366-87577f4cad57

**Add from Code Library button on valueset selection (no custom codes yet)**
https://github.com/user-attachments/assets/9691200e-70f3-4a8b-bdc7-501ba12ffac5

**Add from Code Library button on valueset selection (yes custom codes)**
https://github.com/user-attachments/assets/d13b543c-0d3b-4fb9-8645-42a3ee23e9e8

## Related Issue

Fixes #

## Additional Information

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation

### DIFF
--- a/src/app/(pages)/queryBuilding/querySelection/QueryLibrary.tsx
+++ b/src/app/(pages)/queryBuilding/querySelection/QueryLibrary.tsx
@@ -21,7 +21,7 @@ import {
 import { DataContext, DataContextValue } from "@/app/shared/DataProvider";
 import classNames from "classnames";
 import { getConditionsData } from "@/app/shared/database-service";
-import { ConditionsMap } from "../utils";
+import { ConditionsMap, EMPTY_QUERY_SELECTION } from "../utils";
 import { CustomUserQuery } from "@/app/models/entities/query";
 import {
   CUSTOM_CONDITION_NAME,
@@ -108,7 +108,13 @@ export const MyQueriesDisplay: React.FC<UserQueriesDisplayProps> = ({
         <h1 className="flex-align-center margin-0">Query Library</h1>
         <div className="margin-left-auto">
           <Button
-            onClick={() => setBuildStep("condition")}
+            onClick={() => {
+              setBuildStep("condition");
+              queryContext?.setSelectedQuery?.(
+                structuredClone(EMPTY_QUERY_SELECTION),
+              );
+              queryContext?.setData?.(null);
+            }}
             className={styles.createQueryButton}
             type="button"
           >
--- a/src/app/(pages)/queryBuilding/querySelection/QueryLibrary.tsx
+++ b/src/app/(pages)/queryBuilding/querySelection/QueryLibrary.tsx
@@ -11,7 +11,13 @@ import Table from "@/app/ui/designSystem/table/Table";
 import { ModalRef } from "@/app/ui/designSystem/modal/Modal";
 import styles from "./querySelection.module.scss";
 import { BuildStep } from "@/app/shared/constants";
-import { renderModal, confirmDelete, handleCopy } from "./utils";
+import {
+  renderModal,
+  handleDelete,
+  SelectedQueryDetails,
+  confirmDelete,
+  handleCopy,
+} from "./utils";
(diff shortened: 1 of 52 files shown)


src/app/(pages)/queryBuilding/querySelection/QueryLibrary.tsx
```diff
@@ -32,41 +30,43 @@ import {
...
-      queryName: queryName,
-      queryId: queryId,
-    });
+    if (queryContext.setSelectedQuery) {
+      queryContext.setSelectedQuery({ queryName, queryId });
+    }
     setBuildStep("valueset");
   };
 
+  const [deletedQuery, setdeletedQuery] = useState<SelectedQueryDetails>({
```
fzhao99, line 65: nit: `setdeletedQuery` --> `setDeletedQuery`

src/app/(pages)/queryBuilding/querySelection/QueryLibrary.tsx
```diff
@@ -32,41 +30,43 @@ import {
...
  */
 export const MyQueriesDisplay: React.FC<UserQueriesDisplayProps> = ({
   queries: initialQueries,
-  selectedQuery,
-  setSelectedQuery,
   setBuildStep,
 }) => {
-  const queriesContext = useContext(DataContext);
+  const queryContext = useContext(DataContext);
+  if (!queryContext?.setSelectedQuery) {
```
fzhao99, line 49: instead of throwing here in the body of the component, could we try / catch in the `handleEdit` function and using the toast pattern we have elsewhere? Think in cases where a dev messed up, the only thing that would be broken would be the update which we'd indicate to users via a toast flow.

This feels like a more graceful fallback compared to rendering the general "client side error" page

fzhao99 (commented):
left a few suggestions! great start on this

src/app/(pages)/queryBuilding/querySelection/QueryLibrary.tsx
robertandremitchell: updated!

src/app/shared/DataProvider.tsx
```diff
@@ -49,6 +50,9 @@ export function DataProvider({
   const [toastConfig, setToastConfig] = useState<ToastConfigOptions | null>(
     null,
   );
+  const [selectedQuery, setSelectedQuery] = useState<SelectedQueryDetails>(
+    EMPTY_QUERY_SELECTION,
```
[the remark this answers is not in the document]
robertandremitchell: oh, hrm, interesting. so in some ways it is working as intended, but I also see how this isn't what we'd want in all cases. So in an ideal use case, I think we would want to continue to hold it in memory until they hit the "Create query" button, rather than editing the in-progress query, which should now appear in the query library. So we will want to invoke the `resetQueryState` on `Create query` to get us back to that state. I can work on that.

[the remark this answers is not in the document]
robertandremitchell: @fzhao99 , I added reset logic to the Create query button. I _think_ that should be sufficient, since if they go back to `/queryBuilding` and hit `Edit`, the dataContext won't update, preserving any updates they made on the `/codeLibrary` page. Once it's all tied together, I think we will probably need to stress test, determine if there are other scenarios where we want to trigger reset.

fzhao99 (approved):
lgtm!

katyasoup (approved):

PR #610, opened 2025-05-16
Form validation: check required fields before saving a custom value set

### LINKED ISSUE #504: Add input validation for new value set page
## Description:

e should have some sort of validation on codes when adding new value sets

## Acceptance Criteria:

katyasoup:
# PULL REQUEST

## Summary
Adds a check for the three required fields (Name, Category, System) when saving a new or existing custom value set:
- If blank, the input is outlined in red and an icon with warning text appears for each field; the value set is not saved to the db
- Error styles are removed on input change (entering text/selecting a dropdown option)
- Codes (code # and display/name) can be left blank; the value set will be saved without any code associations

## Related Issue

Fixes #504 

## Additional Information
Since codes are not required, I left them as-is, however we may want some sort of alert/warning that partial code data means the code won't be saved

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation


src/app/(pages)/codeLibrary/components/CustomValueSetForm.tsx
```diff
@@ -334,15 +331,20 @@ const CustomValueSetForm: React.FC<CustomValueSetFormProps> = ({
...
         });
 
-        setCodes(updatedCodeMap);
+        setCodesMap(updatedCodeMap);
       }
     } catch (error) {
-      console.log(error);
+      showToastConfirmation({
+        variant: "error",
+        body: `Unable to dalue set "${customValueSet.valueSetName}". Please try again later.`,
```
fzhao99, line 339: typo here?

fzhao99 (approved):
think I found one typo but otherwise LGTM!

robertandremitchell (commented):
I noticed one bug, which is that the update after an initial creation seems a bit off, in that it creates two valuesets:

https://github.com/user-attachments/assets/edecbd3f-2bce-4180-9030-8782dccd0009

another I noticed on validation is that we can pass a valueset with no codes set. since we generate the valueset ids ourselves, these will functionally be junk for finding any data, right? so we should require at least one code to be present?


https://github.com/user-attachments/assets/020f7fb6-9122-4cb4-8105-a8718f9d6751

katyasoup:
> I noticed one bug, which is that the update after an initial creation seems a bit off, in that it creates two valuesets:

Will take a look!

> another I noticed on validation is that we can pass a valueset with no codes set. since we generate the valueset ids ourselves, these will functionally be junk for finding any data, right? so we should require at least one code to be present? I see your note; I guess it makes sense to potentially still allow since someone could come back later. I guess I would worry a little about users thinking that a salient name of a valueset == code. But maybe that's not something we should expect.

Yeah, I talked to @mikang about this and we landed on allowing code-less value sets, since the user can come back and add to them later. Maybe we want to add something like not having codeless value sets as an option for adding to queries, or showing a warning that there are no codes when you select it? I can add it as a discussion point for the larger group!

robertandremitchell (approved):
lgtm!

PR #615, opened 2025-05-19
Adding warning modal for unsaved query building changes

### LINKED ISSUE #554: When I try navigating to a different page without saving the query, pop up a dialog that asks if I want to save the query
## Acceptance Criteria:

robertandremitchell:
# PULL REQUEST

## Summary

This adds a warning modal to stop unsaved changes on the /queryBuilding page. We will likely want to extend to /codeLibrary hybrid workflow as well, but starting here. But basically we have 3 potential save states to update the query in the DB and the DataContext:

1. User hits `Customize query` or `Save query`. This is an auto-save, no comparison to existing data.
2. User hits `Add from code library` or `start from scratch`. This is an auto-save, no comparison to existing data.
3. User makes edit(s) then hits a back navigation. This triggers a warning modal where it only saves if the user hits save.

I think the one part I can't figure out is triggering the warning modal before a reload/clicking the queryBuilding in the dropdown. Because it's entangled with the header logic, I don't know if on a reload or click away from the page, we do the has unsaved changes check, but I've had trouble thinking of how that would work. If a user is working mid-query, changes their mind, then goes to just build a new query themselves on /codeLibrary, it would be pretty difficult to disambiguate, imo. I think we could just save a header navigation away is an accepted hard reset.

## Related Issue

Fixes #554 

## Additional Information

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation

### DIFF
--- a/src/app/(pages)/codeLibrary/components/__snapshots__/DropdownFilter.test.tsx.snap
+++ b/src/app/(pages)/codeLibrary/components/__snapshots__/DropdownFilter.test.tsx.snap
@@ -214,6 +214,11 @@ exports[`DropdownFilter renders correctly 1`] = `
               QC Admin
             </option>
             <option
+              value="Hogwarts"
+            >
+              Hogwarts
+            </option>
+            <option
               value="My Team(s)"
             >
               My Team(s)


src/app/(pages)/queryBuilding/buildFromTemplates/BuildFromTemplates.tsx
```diff
@@ -96,6 +98,37 @@ const BuildFromTemplates: React.FC<BuildFromTemplatesProps> = ({
...
+
+    if (!savedQueryRef.current) {
+      return Boolean(hasData); // user typed or selected something
+    }
+    return (
+      JSON.stringify(savedQueryRef.current.constructedQuery) !==
+        JSON.stringify(constructedQuery) ||
+      savedQueryRef.current.queryName !== queryName
+    );
+  }
```
fzhao99, line 121: if we wanted to render the modal on page reload / nav away, think we could use this pattern and attach it to the browser API events that fire on page refresh / navigation. 

https://stackoverflow.com/questions/79089312/force-refresh-screen-when-navigating-back-to-previous#:~:text=I've%20come%20across%20a%20solution%20that%20uses,page%20reload.%20Here's%20the%20code%20I'm%20considering:

robertandremitchell: looking today. may be worth a parking lot / eng sync discussion to figure out if this is what we'd want. I'm zigzagging on what I think makes sense for a reload/nav away.

katyasoup: I think it'd be useful to have this on nav away - I can imagine users clicking to go to code library from the menu (despite having a link in the sidebar) or like, clicking over to user management to see who is assigned to the query already, etc. 

A hard reset when refreshing the page is acceptable/expected, IMO

katyasoup (approved):

src/app/(pages)/queryBuilding/buildFromTemplates/BuildFromTemplates.tsx
robertandremitchell: yeah, that makes sense to me. I can revisit tomorrow (time got away from me today 😅 and see if I can get the modal to trigger then)

PR #617, opened 2025-05-21
Initializing Select mode for Code Library page

robertandremitchell:
# PULL REQUEST

## Summary

This adds support for the select state of the page, which can be accessed either by creating a new query and hitting start from scratch, or by an existing saved query and hitting add from code library.

Tests/review can be done by:
1. Navigate to /queryBuilding.
2. Create new query.
3. Give your query a name.
4. Hit start from scratch.
5. Review the /codeLibrary select mode.
6. Make any selections of valuesets.
7. Hit Next: create query.
8. Review selections.
9. Go back to the /codeLibrary select mode by hitting Add from code library.

A few open questions:
* The Figma designs drop the filter for "select" mode. However, I think it could still be handy? But if it looks crowded or may not be necessary, we could drop.
* In what contexts should the "Next: create query" button be disabled? We could have scenarios where a user loads the page, then decides against adding additional codes, for example. I originally had it where if there were no customcodes selected, it would be disabled, but it seemed odd sequencing wise, as there could be times a user navs there, changes their mind. If they have custom codes already, they get the button, and if they don't they're stuck? Also, I feel like the copy for this button doesn't work for the Add from code library workflow. Maybe we make it `Next: review query`?
* When and how should we "delete" custom valuesets? For example, right now if you add a bunch of valuesets, then deselect them on the /queryBuilding version of the page, they just kinda end up vibing there. Do we need to support/implement a drop/delete feature for these? Or on save, "pop" these now empty valuesets from our queryData to avoid having it clogged?
* A bigger backend engineering question: how should we break up manage/select? I had initially tried for a while yesterday, as I did foresee the page.tsx becoming quite massive. Less of an issue for this ticket, but I think we will probably want to find some way to do it while acknowledging that they're like 80-90% similar. I have not written tests in the event we do want to try to disentangle and simplify page.tsx in a subsequent PR.

## Related Issue

Fixes #257 , partially

## Additional Information
* Still investigating why this introduced an error on another test suite...
* I also _much_ prefer how Katie has handled the navigating to/from queryBuilding on her PR in #620. It may make sense to merge that first, so we can avail ourselves of that here.

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation


src/app/(pages)/codeLibrary/codeLibrary.module.scss
```diff
@@ -168,6 +168,16 @@
   }
 }
 
+.manageCodesLink {
+  font-weight: 700;
+  text-decoration: none;
+
+  &:active,
+  &:visited {
+    color: #0076d6;
```
fzhao99, line 177: could we swap this hex code out with the relevant variable (or equivalent) from the new design system? alternatively in the JSX you can just use a `<Button unstyled />` setup

src/app/(pages)/codeLibrary/codeLibrary.module.scss
```diff
@@ -168,6 +168,16 @@
   }
 }
 
+.manageCodesLink {
+  font-weight: 700;
```
fzhao99, line 172: think it may be good for us to make a utility style in `custom-styles.scss` that uses the design system color token instead of the hex value on 177 and that keeps the color unchanged on active / visited and then use those for this / future links. That way, any future JSX that needs the link styled that way doesn't need to make new CSS

src/app/(pages)/codeLibrary/page.tsx
```diff
@@ -82,7 +91,123 @@ const CodeLibrary: React.FC = () => {
...
+
+  const handleValueSetToggle = (vsId: string, checked: boolean) => {
+    setCustomCodeIds((prev) => {
+      const vs = prev[vsId];
+      if (!vs) {
+        // Seed all concepts with the checked value
+        const valueSet = valueSets.find((vs) => vs.valueSetId === vsId);
+        if (!valueSet) return prev;
+        return {
+          ...prev,
```
fzhao99, line 126: instead of spreading this information here, could we update the values directly on the `prev` dictionary since we have a key into that value? Something like

```ts
const valueSet = valueSets.find((vs) => vs.valueSetId === vsId);
valueSet.includeValueSet = checked
valueSet.concepts = valueSet.concepts.map((c) => ({...c, include: checked})
prev[vsId] = valueSet
return prev
```

fzhao99 (commented):
great work on this Rob! left a few suggestions

src/app/(pages)/codeLibrary/codeLibrary.module.scss
robertandremitchell: that makes sense. Using `unstyled`, but still needed the bolding and text decoration bits.

fzhao99 (approved):

PR #625, opened 2025-05-23
configure role information seeding from IdP

fzhao99:
# PULL REQUEST

## Summary
Pulls roll info from our IdP's in response to the conversation we had with Dan about the topic

https://github.com/user-attachments/assets/81792f7c-1f22-4ee5-ac2d-62b5e4f1430e

## Related Issue
Fixes [QUE-278](https://linear.app/skylight-cdc/issue/QUE-278/spike-investigate-idp-role-management) 

Created [a followup](https://linear.app/skylight-cdc/issue/QUE-286/update-documentation-to-include-idp-role-configuration-information) for the extension of the user guide that we'll need to add to instruct users on how to configure their IdP's to include this information. Am working on that in parallel but thought to merge this in first.

## Additional Information
- Updated the seed values in Keycloak to use the default roles / mappings. 
     - The expected roles are `standard` `super-admin` and `admin`. Idea is to scope these just to the application (ie within whatever `query-connector` domain applies to their setup) rather than the entirety of their IdP so things like `admin` don't conflict with other settings
- Refactored our auth callback code on sign in to be more modular at the IdP interface level
   - I recorded an example from Keycloak that uses the updated seed values. If you want to test the the Azure strategy, happy to pair to help you get set up! 
- Built out interfaces for Keycloak and Entra, adding in a pattern that hopefully future IdP's we support can build off
- Factored some common functionality into a shared interface

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation


robertandremitchell (approved):
amazing! looks good to me

PR #639, opened 2025-05-28
add table overflow styling

fzhao99:
# PULL REQUEST

## Summary
A low-hanging fruit styling ticket

Before
image

After
Screenshot 2025-05-28 at 12 24 17 PM

## Related Issue
Fixes [QUE-288](https://linear.app/skylight-cdc/issue/QUE-288/overflow-text-on-the-code-library-page-for-longer-codes)

## Checklist
- [x] Link to relevant issues
- [x] Update documentation

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -5,8 +5,8 @@ DATABASE_URL=postgresql://postgres@localhost:5432/tefca_db
 AUTH_DISABLED=false
 NEXT_PUBLIC_AUTH_PROVIDER=keycloak
 AUTH_SECRET="ido5D/uybeAB3AmMQwn+ubw2zYC4t2h7RJlW2R79598="
-LOCAL_KEYCLOAK=http://localhost:8081
-NAMED_KEYCLOAK=http://localhost:8081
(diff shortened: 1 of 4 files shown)


robertandremitchell (commented):
makes sense, I thought I had fixed this, but I had done it in the `select` view of the page where it wraps to the next line, which I think we also do with long codes in the `valuesets` view on the `/queryBuilding` page. I don't have a strongly held view on how to handle long codes, but they should probably be in alignment imo.

fzhao99:
> makes sense, I thought I had fixed this, but I had done it in the `select` view of the page 

I think these are wrapped in the `Highlighter` component, which sticks a couple extra HTML elements that's throwing off the styling that you might have added.


>where it wraps to the next line, which I think we also do with long codes in the `valuesets` view on the `/queryBuilding` page. I don't have a strongly held view on how to handle long codes, but they should probably be in alignment imo.

I'll loop in Michelle and ask which she prefers!

katyasoup (approved):

PR #640, opened 2025-05-28
fix audit log null pointer

fzhao99:
# PULL REQUEST

## Summary
Fixing a null pointer Dan pointed out [here](https://skylight-hq.slack.com/archives/C05BEG226RZ/p1748449284560689)

A little hard to debug here since we're dealing with environment mismatch issues but based on this console log, think this is the source of the error. Will verify this fixes it once it deploys
Screenshot 2025-05-28 at 12 35 10 PM

## Checklist
- [x] Link to relevant issues


robertandremitchell (approved):
makes sense

PR #647, opened 2025-05-28
Prefill patient search from URL params

### LINKED ISSUE #642: Pre-fill patient search from URL params
## Acceptance Criteria:

katyasoup:
# PULL REQUEST

## Summary
When present, values from URL parameters will pre-fill the applicable form fields on the patient search page 
(valid params are `first`, `last`, `phone`, `street`, `street2`, `city`, `state`, `zip`, `dob`, `mrn` & `server`)

## Related Issue

Fixes #642 

## Additional Information
- ~Did not include address since we seem to not be doing anything with that field currently?~ _Updated to pre-fill all address fields, state must match on list and zip must pass regex validation_
- All fields except `server` pull whatever is typed in the URL
  - I added an additional check for `server` so that it only prefills if the query param string matches the list of available servers. Since the server options are hidden by default, and a bad server name breaks the request, this makes it a bit more user-friendly in case of typos, etc.

## Checklist

- [x] Link to relevant issues
- [x] Update documentation


nickclyde (approved):
Looks great, thanks Katie!!

PR #671, opened 2025-06-02
Update endpoint for testing fhir server connections

### LINKED ISSUE #670: Allow empty patient identifiers when testing server connections
## Description:

We recently added restrictions that require a request include a minimum number of patient identifiers, however this cause the "Test connection" functionality in the FHIR server config page to always return an error

## Acceptance Criteria:

* "Test connection" doesn't require patient identifiers for a succesful call
* Patient identifiers are still required for actual patient queries

katyasoup:
# PULL REQUEST

## Summary
Updates the "test connection" on the FHIR server config page to check the `/metadata` endpoint, rather than `/Patient`. This endpoint returns a [Capability Statement](https://hl7.org/fhir/http.html#capabilities) that does not require Patient identifiers to be present on the request.

## Related Issue

Fixes #670 

## Additional Information

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation


nickclyde:
Okay so this is a tricky one, because we want to be able to test a proper authentication with "Test connection", but I think typically `/metadata` is an unauthenticated endpoint. Let me look around to see if there is a better option of an endpoint to hit, open to suggestions as well.

fzhao99 (approved):

katyasoup:
@nickclyde @fzhao99 I added back the `/Patient` endpoint, but threw in a dummy string for the name param - expected result is bundle response with 0 resources. Should all be working now!

nickclyde (approved):

PR #701, opened 2025-06-16
Expanding date range picker to support timeboxing

### LINKED ISSUE #700: Update Date Range Picker to have preset date ranges.

robertandremitchell:
# PULL REQUEST

## Summary

This expands the date range picker to support timeboxing. It is currently viewable and testable on the /auditLogs. The bulk of this should then be reusable for timeboxing on the queryBuilding page, once the backend is implemented (separate ticket). Based on the design here:

image

https://github.com/user-attachments/assets/edf67abe-1879-489d-bb04-6c4a20666d74

## Related Issue

Fixes #700 

## Additional Information

## Checklist

- [ ] Link to relevant issues
- [ ] Update documentation

### DIFF
--- a/e2e/custom_headers.spec.ts
+++ b/e2e/custom_headers.spec.ts
@@ -1,0 +1,326 @@
+import { TEST_URL } from "../playwright-setup";
+import { test, expect } from "@playwright/test";
+import { E2E_SMART_TEST_CLIENT_ID } from "./constants";
+
+test.describe("Custom Headers", () => {
+  test("successfully adds a FHIR server with custom headers", async ({
+    page,
+  }) => {
+    await page.goto(`${TEST_URL}/fhirServers`);
+    await expect(
+      page.getByRole("heading", { name: "FHIR server configuration" }),
+    ).toBeVisible();
+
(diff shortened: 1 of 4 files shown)


src/app/ui/designSystem/timeboxing/DateRangePicker.tsx
```diff
@@ -31,8 +33,103 @@ const normalizeStart = (date: Date | null) =>
 const normalizeEnd = (date: Date | null) =>
   date ? new Date(date.setHours(23, 59, 59, 999)) : null;
```
fzhao99, line 34: do we want to do any normailzation of the time zone here as well? Think by default the Date uses the system timezone, which may result in things getting added / excluded from the filter unexpectedly

robertandremitchell: Removed the normalization code from 24 hours, so I think that should cover that case, if we do want that to be as explicit as possible. I guess my reason for not doing so initially is that even within a state (or county/town!) there can be 2 timezones, but that's such an edgy case idk if it's better to be technically correct or more lenient with the range.

PR #702, opened 2025-06-16
Add ability to configure FHIR server headers in frontend

### LINKED ISSUE #653: Add support for setting headers for FHIR servers
## Description:

Using eHX servers requires us to send headers with all requests, but right now we only support that by putting them directly into the DB. We should add some fields to the modal to support adding headers

## Acceptance Criteria:

nickclyde:
# PULL REQUEST

## Summary

Adds the ability to configure headers for FHIR servers. Can add, edit, remove each key value pair independently. Tried to make it clear too that the Authorization header is handled separately.

Screenshot 2025-06-16 at 4 06 13 PM

Screenshot 2025-06-16 at 4 06 58 PM

## Related Issue

Fixes #653 (https://linear.app/skylight-cdc/issue/QUE-294/add-support-for-setting-headers-for-fhir-servers)

## Copilot summary

This pull request introduces support for managing custom HTTP headers in the FHIR server configuration, including both frontend and backend changes. The changes enable users to add, update, and remove custom headers through the UI, and ensure these headers are properly handled during server interactions and persisted in the database. Additionally, integration tests have been added to validate the functionality.

### Frontend Changes

* Added a `HeaderPair` interface and a `headers` state to manage custom headers in the `FhirServers` component. This includes methods to add, update, and remove headers, as well as a utility function to convert the header state into an object. (`src/app/(pages)/fhirServers/page.tsx`, [[1]](diffhunk://#diff-da48056c7881d1c27c453825ad7bfdaac4227afc8aa4c4aeac36a4c3150f5535R37-R42) [[2]](diffhunk://#diff-da48056c7881d1c27c453825ad7bfdaac4227afc8aa4c4aeac36a4c3150f5535R68) [[3]](diffhunk://#diff-da48056c7881d1c27c453825ad7bfdaac4227afc8aa4c4aeac36a4c3150f5535R190-R218)
* Updated the modal UI to include a section for managing custom headers, with input fields for header keys and values, and buttons for adding or removing headers. (`src/app/(pages)/fhirServers/page.tsx`, [src/app/(pages)/fhirServers/page.tsxR722-R778](diffhunk://#diff-da48056c7881d1c27c453825ad7bfdaac4227afc8aa4c4aeac36a4c3150f5535R722-R778))
* Modified the "Test connection" and "Save" functionalities to include custom headers in the request payload. (`src/app/(pages)/fhirServers/page.tsx`, [[1]](diffhunk://#diff-da48056c7881d1c27c453825ad7bfdaac4227afc8aa4c4aeac36a4c3150f5535L180-R231) [[2]](diffhunk://#diff-da48056c7881d1c27c453825ad7bfdaac4227afc8aa4c4aeac36a4c3150f5535R304-R311)

### Backend Changes

* Updated the `FhirServerConfigService` to handle custom headers during server creation and updates, ensuring that the `Authorization` header is excluded from user-defined headers and managed separately based on the authentication method. (`src/app/backend/fhir-servers.ts`, [[1]](diffhunk://#diff-bea67cc8a8de50cbf418de733c8ffe88e52ba9893ae2810ee7bcbbb37055f6b3L159-L176) [[2]](diffhunk://#diff-bea67cc8a8de50cbf418de733c8ffe88e52ba9893ae2810ee7bcbbb37055f6b3L251-R255)

### Testing Enhancements

* Added integration tests to verify the functionality of custom headers, including scenarios for inserting, updating, and preserving headers, as well as handling headers with different authentication methods (e.g., basic, client credentials, SMART). (`src/app/tests/integration/fhir-servers.test.ts`, [src/app/tests/integration/fhir-servers.test.tsR94-R380](diffhunk://#diff-dd7b01c0b86dd1a7e0fd5c450390de9074e8c1f213960c40aa30f392b73bf7beR94-R380))

### DIFF
--- a/src/app/(pages)/fhirServers/page.tsx
+++ b/src/app/(pages)/fhirServers/page.tsx
@@ -34,6 +34,12 @@ const Modal = dynamic<ModalProps>(
 type ModalMode = "create" | "edit";
 type AuthMethodType = "none" | "basic" | "client_credentials" | "SMART";
 
+interface HeaderPair {
+  key: string;
+  value: string;
+  id: string; // For React key prop
+}
+
 /**
  * Client side parent component for the FHIR servers page. It displays a list of FHIR servers
  * @returns - The FhirServers component.
@@ -59,6 +65,7 @@ const FhirServers: React.FC = () => {
     null,
   );
   const [modalMode, setModalMode] = useState<ModalMode>("create");
+  const [headers, setHeaders] = useState<HeaderPair[]>([]);
   const modalRef = useRef<ModalRef>(null);
 
   // Fetch FHIR servers
@@ -93,6 +100,7 @@ const FhirServers: React.FC = () => {
     setDefaultServer(false);
     setErrorMessage("");
     setSelectedServer(null);
+    setHeaders([]);
   };
 
   const handleOpenModal = (mode: ModalMode, server?: FhirServerConfig) => {
@@ -105,6 +113,20 @@ const FhirServers: React.FC = () => {
       setDisableCertValidation(server.disableCertValidation);
       setDefaultServer(server.defaultServer);
       setErrorMessage("");
+
+      // Set headers
+      if (server.headers) {
+        const headerPairs: HeaderPair[] = Object.entries(server.headers)
+          .filter(([key]) => key !== "Authorization") // Filter out Authorization header
+          .map(([key, value]) => ({
+            key,
+            value,
+            id: `${Date.now()}-${Math.random()}`,
+          }));
+        setHeaders(headerPairs);
+      } else {
+        setHeaders([]);
+      }
 
       // Set auth method and corresponding fields based on server data
       if (server.authType) {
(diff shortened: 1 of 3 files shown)


src/app/(pages)/fhirServers/page.tsx
```diff
@@ -664,6 +719,63 @@ const FhirServers: React.FC = () => {
...
+                  />
+                </div>
+                <div className="grid-col margin-left-2">
+                  <button
+                    type="button"
+                    className="usa-button usa-button--unstyled margin-top-2"
+                    onClick={() => removeHeader(header.id)}
+                    aria-label={`Remove header ${header.key || "row"}`}
+                  >
+                    <Icon.Close size={3} />
```
robertandremitchell, line 763: these have a blue `X` instead of the default purple `X`.

src/app/(pages)/fhirServers/page.tsx
```diff
@@ -165,6 +187,35 @@ const FhirServers: React.FC = () => {
...
+        header.id === id ? { ...header, [field]: value } : header,
+      ),
+    );
+  };
+
+  const removeHeader = (id: string) => {
+    setHeaders(headers.filter((header) => header.id !== id));
+  };
+
+  const convertHeadersToObject = (): Record<string, string> => {
```
fzhao99, line 209: just for my understanding, what's the advantage of storing the headers as an array and then normalizing it into an object rather than just storing the headers as an object? I think it's because it makes the appending process for additional headers easier (since we don't have to validate the keys in the final dict until we're ready to save) but wanted to check my understanding

nickclyde: Correct - it was just easier, especially since I'm using `Object.entries` to parse the headers object stored in the DB, and that returns an array of arrays, it was easier to just map that to an array of HeaderPair objects, which could also store the unique ID for the React key prop, which we could then throw away when converting back to an object. Also deleting from arrays is much easier than deleting keys from objects!

PR #730, opened 2025-07-14
split apart value set selection

fzhao99:
# PULL REQUEST

## Summary
As part of the timeboxing work, noticed ValueSetSelection component was growing a bit unwieldy, so I went ahead and broke it apart into smaller views:
- A view for the `Sidebar`, which also has drawer logic in it for the condition maps
- A view for custom codes
- A view for Medical records

This helps to align the level of abstraction in the parent `ValueSetSelection.tsx` file to organizing / arranging the major views on the page (sidebar + 3 views that are in that sidebar) and hopefully makes the page more scalable

In the process, also made the drawer open / close dependent on Javascript rather than CSS to better represent the state of the UI / to make testing easier. Changed some tests / snapshots as a result.

## Additional Information

## Checklist
- [x] Link to relevant issues


robertandremitchell (commented):
looks good to me! definitely reads more clearly

robertandremitchell (approved):

PR #757, opened 2025-07-28
add a regression test

fzhao99:
# PULL REQUEST

## Summary
It seemed that we had a regression for query filtering, so added that functionality back in and added a regression e2e since it seems like a pretty core piece of functionality

## Checklist


robertandremitchell (approved):
lgtm!

PR #793, opened 2025-08-13
make time filtering relative

fzhao99:
# PULL REQUEST

## Summary
After a conversation with Dan, adjusting the time filtering to be relative by default. The radio buttons should now be relative to the current date, while the custom date range still have the ability to be absolute.

In the recording below, logged out the time stamps for a default query: notice how the seconds values update between the first and second query
https://github.com/user-attachments/assets/ac5ad95d-8b3d-49c5-a031-d98ed242dad2

Think for now, calling it good to just having the preset options / ability to make an absolute time range. Will probably need some design input for the component to extend to arbitrary time ranges / units.
 
## Checklist


nickclyde (approved):
Looks great, thanks Bob!

PR #831, opened 2025-09-08
Add CA Certificate field to FHIR servers

### LINKED ISSUE #779: Add trust for the eHealthExchange certificate authority
## Description:

To remove dependency on insecure fetch, make the cert authority from eHealth a trusted root cert in the service

## Acceptance Criteria:

nickclyde:
# PULL REQUEST

## Summary

Adds a field to FHIR server config for CA certificates, only visible when Mutual TLS is selected for auth. This CA cert gets passed to the custom fetch method for the client, enabling mutual TLS with certificates generated by a non-default CA.

## Related Issue

Fixes #779 

Screenshot 2025-09-08 at 3 19 33 PM


src/app/backend/fhir-servers/fhir-client.ts
```diff
@@ -58,9 +61,12 @@ class FHIRClient {
       try {
         const cert = getOrCreateMtlsCert();
         const key = getOrCreateMtlsKey();
+        const ca = config.caCert || "";
+        console.log(`CA cert length: ${ca.length}`);
```
fzhao99, line 65: did we mean to pull this?

src/app/backend/db/util.ts
```diff
@@ -14,9 +14,10 @@ INSERT INTO fhir_servers (
   scopes,
   access_token,
   token_expiry,
+  ca_cert,
```
fzhao99, line 17: do we want to base64 encode these values when we're storing them? ran into issues with the newlines in CI when I was dealing with the variables, so wondering if it'll be simpler to store them as contiguous strings and then unencode them in the interface body when we send off the request

nickclyde: Since they're just being stored in the database, I don't think we'll run into the same issue, right? For tests, we can just use template strings hardcoded in the test and I think it should work okay. But I guess I should get the tests passing properly before I decide if it works 😅

src/app/(pages)/fhirServers/fhirServersModal.tsx
```diff
@@ -659,9 +664,41 @@ export const FhirServersModal: React.FC<FhirServersModal> = ({
...
             <div className="usa-hint margin-top-05">
-              Mutual TLS certificates will be loaded from the keys directory or
-              MTLS_CERT and MTLS_KEY environment variables.
+              Mutual TLS client certificates will be loaded from the keys
+              directory or MTLS_CERT and MTLS_KEY environment variables. Provide
+              the CA certificate of the server here.
             </div>
+            <Label htmlFor="ca-cert">
+              Server CA Certificate{" "}
+              <span className="text-secondary">(required)</span>
```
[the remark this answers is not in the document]
nickclyde: Not if the server's cert is issued by a CA that's trusted by default, true. So you're right, it shouldn't necessarily be required. Maybe a nice enhancement in the future would be to automatically grab the cert and populate in when the user clicks "Test connection" 🤔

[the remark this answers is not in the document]
nickclyde: Made optional!

PR #922, opened 2026-04-23
Fix DB secret rotation and catch special chars in password

nickclyde:
Attempt to fix DB rotation for Colorado and throw an error for db passwords with special characters


nickclyde:
Closing for now.

